PR #129, opened 2022-05-19
add: magic world filter

patrykcieszkowski:
adds location filter, behaving like no filter at all.  ¯\(°_o)/¯

```json
{ "type": "magic", "value": "world", "limit": 2 }
```


patrykcieszkowski:
@jimaek

jimaek:
In the name of good UX I pronounce this PR merged

PR #143, opened 2022-06-17
feat: nested magic filters

patrykcieszkowski:
- [ ] router logic
- [ ] unit tests

### DIFF
--- a/src/probe/router.ts
+++ b/src/probe/router.ts
@@ -39,7 +39,9 @@ export class ProbeRouter {
 				return this.filterGloballyDistributed(sockets, sockets.length);
 			}
 
-			return sockets.filter(s => s.data.probe.index.find(v => v.includes(location.value.replace('-', ' ').toLowerCase())));
+			const locationList = location.value.split('+').map(l => l.replace('-', ' ').toLowerCase());
+
+			return sockets.filter(s => locationList.every(l => s.data.probe.index.includes(l)));
 		}
 
 		return sockets.filter(s => s.data.probe.location[location.type] === location.value);


jimaek:
Lets also document this. I think we released some other features as well that weren't documented.

patrykcieszkowski:
@zarianec any idea why these tests fail? Increasing timeout on `Create measurement` does the job, but that doesn't seem right.

docs/measurement/schema/location.md
```diff
@@ -176,3 +177,9 @@ both of the following queries would match `amazon technologies inc.` network
 { "type": "magic", "value": "aws" },
 { "type": "magic", "value": "amazon" }
 ```
+
+magic queries can be combined. The following query will match `London, Ohio, USA`
+
+```json
+{ "type": "magic", "value": "london+ohio+usa" }
```
jimaek, line 184: In our case it wont match anything since we dont have a probe in London in Ohio.
Maybe a clearer example that would work on our existing probes? E.g. "google+belgium" to return results from Google Cloud probes only in Belgium.

patrykcieszkowski: I wanted to show you can form a very precise queries, but I guess this exmaple would be close enough.

jimaek: I understand but I want our examples to be copy-paste functioning :)

patrykcieszkowski:
I re-run the check and it passed just fine. I suppose it must've been caused by some overload on GH servers.

Either way, we use 15s timeout on other HTTP tests, so I applied the same logic to this test, just in case.

PR #144, opened 2022-06-20
feat: nested filters

patrykcieszkowski:
https://github.com/jsdelivr/globalping/pull/143 should be merged first.

- [x] probe router logic
- [x] schema
- [x] error message
- [x] unit tests
- [x] demo page
- [x] update magic match (requires #143 merge)
- [x] docs

rules:
- `filter: combined | default` has to be added on 0-level body (next to global limit)
- `filter: default` is ignored
- `filter: combined` can not be used with per-location limits


patrykcieszkowski:
@jimaek

PR #175, opened 2022-07-26
feat: critical version check

patrykcieszkowski:
relies on https://github.com/jsdelivr/globalping-probe/pull/85


src/probe/builder.ts
```diff
@@ -42,7 +43,7 @@ export const buildProbe = async (socket: Socket): Promise<Probe> => {
 		throw new Error('failed to detect ip address of connected probe');
 	}
 
-	if (!VERSION_REG_EXP.test(version)) {
+	if (!VERSION_REG_EXP.test(version) || !semver.satisfies(version, '^0.8.1')) {
```
MartinKolarik, line 46: We can remove the regex test because semver checks for validity too.

patrykcieszkowski: done.

PR #180, opened 2022-07-27
chore: add tsx as another npm script

ayuhito:
Adds `tsx` as an alternative dev script. Skips the need for `tsc` which was taking me 20-30s on my machine. Now takes under a second to compile and run 🎉 

Not sure if there's interest in replacing the current `dev` script with this, so I used a separate name.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "rimraf": "^3.0.2",
     "sinon": "^14.0.0",
     "supertest": "^6.2.2",
-    "ts-node": "^10.7.0",
     "tsx": "^3.8.0",
     "typescript": "^4.6.4",
     "xo": "^0.49.0"
@@ -87,8 +86,10 @@
     "dev:tsx": "NODE_ENV=development tsx src/index.ts",
     "lint": "xo",
     "lint:fix": "xo --fix",
+    "lint:types": "tsc --noEmit",
     "blacklist": "node dist/lib/malware/download.js",
-    "test": "NODE_ENV=test mocha",
+    "test": "npm run lint:types && npm run test:src",
+    "test:src": "NODE_ENV=test mocha",
     "clean": "rimraf coverage",
     "test:coverage": "npm run clean && c8 mocha",
     "init:hooks": "husky install"
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "sinon": "^14.0.0",
     "supertest": "^6.2.2",
     "ts-node": "^10.7.0",
+    "tsx": "^3.8.0",
     "typescript": "^4.6.4",
     "xo": "^0.49.0"
   },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -83,6 +84,7 @@
     "build": "tsc && npm run blacklist",
     "start": "NODE_ENV=production node dist/index.js",
     "dev": "NODE_ENV=development node dist/index.js",
+    "tsx": "NODE_ENV=development tsx src/index.ts",
```
patrykcieszkowski, line 87: let's name it `dev:tsx`, since its only used for dev env.

patrykcieszkowski (approved):

ayuhito:
@patrykcieszkowski, what are your thoughts on switching to `loader=tsx`
https://github.com/jsdelivr/globalping/blob/133b45b6ce2de4f1fb1540c5adcc19be71211da9/.mocharc.json#L10

Speeds things significantly on my end when running tests, since I think ts-node still uses tsc under the hood.

patrykcieszkowski:
> @patrykcieszkowski, what are your thoughts on switching to `loader=tsx`
> 
> https://github.com/jsdelivr/globalping/blob/133b45b6ce2de4f1fb1540c5adcc19be71211da9/.mocharc.json#L10
> 
> Speeds things significantly on my end when running tests, since I think ts-node still uses tsc under the hood. Also adds caching so we don't recompile untouched test files in subsequent runs.
> 
> Maybe we can just remove the ts-node dependency with a change like that? Or are we using ts-node elsewhere too?

TSX doesn't support type checking, and that's crutial for test to cover type checks as well. Unless we were to run type-checker seperately, but I personally would like to be aware of my type-conflicts IRL. This is also why I'd like to keep `dev` and `dev:tsx` seperately.

ayuhito:
Would you be open to changing `npm test` to `npm run lint:types && npm run test:src` instead? `lint:types` being `tsc --noEmit` and `test:src` being `mocha` with the tsx loader. 

Shouldn't change yours or CI workflow much but lets me (or others) directly skip type-checking if small iterative changes are being made quickly over and over again. My IDE tends to scream at me a lot anyways for type errors so I tend to usually let CI take care of proper type-checking.

patrykcieszkowski:
> Would you be open to changing `npm test` to `npm run lint:types && npm run test:src` instead? `lint:types` being `tsc --noEmit` and `test:src` being `mocha` with the tsx loader.
> 
> Shouldn't change yours or CI workflow much but lets me (or others) directly skip type-checking if small iterative changes are being made quickly over and over again. My IDE tends to scream at me a lot anyways for type errors so I tend to usually let CI take care of proper type-checking.

Sounds good, as long as we have both; types and actual tests covered I see no reason not to do that.

patrykcieszkowski:
Not really an issue, just want to point this out. This change has in fact increased the test time:

pre-change:
```
real    0m8.164s
user    0m12.820s
sys     0m0.776s
```

post-change:
```
real    0m16.380s
user    0m23.101s
sys     0m1.651s
```

patrykcieszkowski (approved):

PR #181, opened 2022-07-27
fix: add ip support to dns ptr

### LINKED ISSUE #172: DNS PTR is useless without IP support
We have PTR as a listed query type but for it to work we need to:
1. Allow IPs as target, only domains are allowed for now
2. Do `dig -x` when PTR is selected

ayuhito:
Partially resolves #172. It adds conditional validation when PTR is selected to prevent domains from being used and only IPs. 

Adds 3 test cases, but I still need to test the whole build locally once I figure out how to do that properly. So might need to wait a little before merging just to be sure.

Do we want IPv6 support for PTR? Currently, globalIpOptions only allows IPv4 but I don't know if that's applicable for this.

I'll look into adding `dig -x` support in the probe repo tomorrow.


jimaek:
No ipv6 support:)

patrykcieszkowski (approved):

MartinKolarik (approved):
👌

PR #182, opened 2022-07-27
docs: update setup instructions

ayuhito:
Add docs on setting up probe for testing locally. Is there anything else that I missed?


ayuhito:
Actually there is one thing that also made setup a little difficult for me. I have `tls: false` set for my local Redis instance using Docker Compose. Otherwise, it'll refuse to connect.
https://github.com/jsdelivr/globalping/blob/133b45b6ce2de4f1fb1540c5adcc19be71211da9/config/default.cjs#L6

Not sure if this issue has been encountered by anyone else? I could include a note on disabling TLS as a possible troubleshooting solution in setup.

patrykcieszkowski:
> Actually there is one thing that also made setup a little difficult for me. I have `tls: false` set for my local Redis instance using Docker Compose. Otherwise, it'll refuse to connect.
> 
> https://github.com/jsdelivr/globalping/blob/133b45b6ce2de4f1fb1540c5adcc19be71211da9/config/default.cjs#L6
> 
> Not sure if this issue has been encountered by anyone else? I could include a note on disabling TLS as a possible troubleshooting solution in setup.



> Actually there is one thing that also made setup a little difficult for me. I have `tls: false` set for my local Redis instance using Docker Compose. Otherwise, it'll refuse to connect.
> 
> https://github.com/jsdelivr/globalping/blob/133b45b6ce2de4f1fb1540c5adcc19be71211da9/config/default.cjs#L6
> 
> Not sure if this issue has been encountered by anyone else? I could include a note on disabling TLS as a possible troubleshooting solution in setup.

That's strange, the redis image we use disables TLS by default.

MartinKolarik:
> > Actually there is one thing that also made setup a little difficult for me. I have `tls: false` set for my local Redis instance using Docker Compose. Otherwise, it'll refuse to connect.
> 
> That's strange, the redis image we use disables TLS by default.

But in our default app config, it's enabled, that's the problem. If redis defaults to disabled, we should probably set `redis.socker.tls` to `false` in the `development.json`

ayuhito:
Should I make changes to `development.cjs` in this PR then?

MartinKolarik:
Sure

config/development.cjs
```diff
@@ -1 +1,7 @@
-module.exports = {};
+module.exports = {
+	redis: {
+		socket: {
+			tls: true,
```
MartinKolarik, line 4: It should be `false` right?

ayuhito: Oops, I'm way too sleepy.

PR #184, opened 2022-07-28
ref: prettify region name

### LINKED ISSUE #179: Update region names
Missed part of #162, "southern america" should be "South America"

patrykcieszkowski:
resolve #179 

- [x] geoip: add `region` details
- [x] add: `normalizedRegion` field
- [x] prettify `region` field
- [x] update region name schema
- [x] tests


src/probe/route/get-probes.ts
```diff
@@ -15,7 +15,17 @@ const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & R
 	ctx.body = socketList.map((socket: Socket) => ({
 		version: socket.data.probe.version,
 		ready: socket.data.probe.ready,
-		location: socket.data.probe.location,
+		location: {
```
MartinKolarik, line 18: Looks like we're missing proper tests for this endpoint.

PR #191, opened 2022-07-31
Create INFRASTRUCTURE.md

jimaek:


INFRASTRUCTURE.md
```diff
@@ -0,0 +1,46 @@
...
+The purpose is to document everything for ourselves as well as allow everyone else to explore our services and contribute their own ideas for potential optimizations.
+
+### API - api.globalping.io
+
+The main production API that all probes and users connect to.
+
+- Hosted on Heroku in Europe 
+- Manually triggered deployments from GitHub `master` branch
+- Integrated into GitHub
+- Single Dyno "Standard 2X"
```
patrykcieszkowski, line 13: why don't we list resource info? 1024GB of RAM and 2 vCPUs.

jimaek: I didn't because with Dynos its not super clear, but I added it now. If we switch to own servers we will definitely have detailed resources

patrykcieszkowski (approved):

PR #193, opened 2022-08-03
ref: redis: unecessary isolated calls + ioredis

### LINKED ISSUE #192: Refactor measurement store module
Spawning new redis connection per measurement update is extremaly inefficient, with little to no benefit.

https://github.com/jsdelivr/globalping/blob/master/src/measurement/store.ts#L31

patrykcieszkowski:
resolve #192 

- [x] lib/Redis
- [x] lib/ws
- [x] lib/cache
- [x] measurement/store
- [x] measurement/runner
- [x] lib/metrics


MartinKolarik (commented):
Considering there doesn't seem to be any benefit from changing the package, not even in benchmarks, let's keep `redis` and update this to only remove the isolated calls + use `Promise.all()`

Issue #163, opened 2022-07-15
Consider using a local DB to manage city name inconsistencies


MartinKolarik:
In some cases, different providers return different variants of the same city name. #162 adds basic normalization for "city" and "the" cases, but overall a more effective solution might need a local DB with a list of all known names used to resolve the correct name. A possible source of the data: https://www.kaggle.com/datasets/geonames/geonames-database?select=geonames.csv

This is not something to do right now but a possible future improvement.

ayuhito:
Alternate datasets: http://www.geonames.org or https://www.getty.edu/research/tools/vocabularies/tgn/index.html

Getty has a much more permissive free license than Geonames. But Geonames is much more comprehensive (although I haven't fully verified the quality of the free license).

Both are actively maintained though.

MartinKolarik:
The file I linked and essentially everything else I found was based on geonames, so yes, that's probably the ultimate source.

ayuhito:
Note that since both sets use licenses that require attribution (CC 4.0 or ODC 1.0), we can use the README or personally I do like https://discord.com/licenses giving credit to everyone in general (and also helps partially audit the licenses we use). In this example, Discord attributes the caniuse's CC 4.0 license.

They also have a fun little popup if you scroll down to the OS libraries tab here: https://discord.com/open-source

ayuhito:
I can give this a go if that's fine.

I was thinking I'll make a separate NPM package called normalise-location that updates itself every month with GH Actions and processes all the data to a usable format. That way we're not adding much more complexity here.

jimaek:
Why would we need frequent updates? Cities rarely change names. We should use an existing source that will minimize the maintenance needed on our side

ayuhito:
It'll be automated in the normalise-location repo and a "whatever" thing to update here like any other dep patch update. The data is fixed otherwise.

https://www.geonames.org/recent-changes.html - They do seem to regularly update their dataset daily (although almost all of it is not relevant), so I think a monthly automated GH Action to rerun the data processing scripts is all that's needed to maintain itself. If there's no change to the data relevant to us, it won't publish a new NPM version.

It's a little overkill but new towns and places still do get added infrequently. And since I'm going to have to make the scripts once anyways, adding an automated GH action to auto-update isn't much extra work.

MartinKolarik:
@ayuhito let's try and see how it works.

ayuhito:
[...]

PR #194, opened 2022-08-03
WIP: fix: use ascii names and custom map

### LINKED ISSUE #190: Remove non-ascii symbols from API
Reminder to follow up on the Slack discussion. I think everyone agrees it only creates problems and adds no value to have them.
Uppercase remains, and we can do normalization stuff to support better English names.

ayuhito:
Closes #190 and #163.

I removed the city element from the Regex as I'm not sure it's valid to denote places like Kansas City or Mexico City as other names. But I guess the same logic could be applied to New York as that can also be considered a state just like Kansas. Thoughts?

I think we should reopen #163 if we think maintaining a manual map is actually taking too much developer time in the long run, which I highly highly doubt. Otherwise, this should be suitable. 

This also fixes a bug with location filtering. While the probe on the demo/api results shows 'Kansas City', you can't use 'Kansas City' as the filter but only 'Kansas'. This makes both what's shown to the user and what should be inputted to be the same and unambiguous.

### DIFF
--- a/src/lib/geoip/altnames.ts
+++ b/src/lib/geoip/altnames.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/naming-convention, quote-props */
-export const altNames: Record<string, string> = {
+export const cities: Record<string, string> = {
 	'Geneve': 'Geneva',
 	'Frankfurt am Main': 'Frankfurt',
 	'New York City': 'New York',
--- a/src/lib/geoip/altnames.ts
+++ b/src/lib/geoip/altnames.ts
@@ -1,0 +1,6 @@
+/* eslint-disable @typescript-eslint/naming-convention, quote-props */
+export const altNames: Record<string, string> = {
+	'Geneve': 'Geneva',
+	'Frankfurt am Main': 'Frankfurt',
+	'New York City': 'New York',
+};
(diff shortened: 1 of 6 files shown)


src/lib/geoip/utils.ts
```diff
@@ -1,5 +1,20 @@
 import anyAscii from 'any-ascii';
 
-export const normalizeCityName = (name: string): string => anyAscii(name).toLowerCase().replace(/(?:\s+|^)(?:the|city)(?:\s+|$)/gi, '');
+/* eslint-disable @typescript-eslint/naming-convention */
+const altNames: Record<string, string> = {
```
patrykcieszkowski, line 4: we should move aliases to a seperate file. Just like we did for ISP names.

https://github.com/jsdelivr/globalping/blob/master/src/lib/location/networks.ts

src/lib/geoip/altnames.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable @typescript-eslint/naming-convention, quote-props */
+export const altNames: Record<string, string> = {
```
patrykcieszkowski, line 2: The variable should be more descriptive; the file name is rather general and it could be assumed it contains aliases for more than just one field.

ayuhito: How does cityAltnames sound? Or do we want to change the filename itself?

patrykcieszkowski: I would name it “cities”, but “cityAltnames” works just fine too.

patrykcieszkowski (approved):

MartinKolarik:
Waiting for resolving discussion in #163.

Issue #200, opened 2022-08-13
Internal metrics collection


jimaek:
For tasks like https://github.com/jsdelivr/globalping/issues/37 and many others we will need to know more about the probes.

Here is a list of data we would probably need to collect to build the next features:

- Accepted tests, in progress tests, finished tests, failed tests. Accepted/in progress probably just real-time value, while the rest are timeseries. Maybe even total and per type as well?
- CPU load and CPU cores available
- Uptime
- Other?

We should probably store them for up to 7 days in a timeseries DB. But note that whatever DB we choose it will need it to scale as in the future we will also support scheduled tests, like pinging the same target every minute and building a chart of performance over time per region.

patrykcieszkowski:
How often would we collect CPU/mem stats? And how often would we pushe them to the API?

Also, how exactly do we define uptime? How long has the probe been connected, or what do we do if the probe reconnects?

jimaek:
> How often would we collect CPU/mem stats? And how often would we push them to the API?

Ideally every few seconds, e.g. every 10s. The more accurate is the data the more we can later do with tests routing between probes. No local buffering, collect and push immediately. 

> Also, how exactly do we define uptime? How long has the probe been connected, or what do we do if the probe reconnects?

Probably how long probe was connected unless there is a better metric. We can technically measure this both from a probe and the API. So in this case its probably better to collect this data on the API level?

If the probe disconnected and connected with "ready" state after 5 seconds this means the probe had a downtime of 5s.

patrykcieszkowski:
@MartinKolarik any idea how to reasonably store this data? Ideally we would keep it all under a single record, but since we need older data to expire, we can't do that. I'm not convinced storing each measurement seperately is a good idea either. Maybe we could group them by 24hr periods, per key?

```
gp:probe:stats:15-08-22
{
    "cpu": [ { "date": "123", ... }, ... ],
    "mem": [ { "date": "123", ... }, ... ]
}
```

I'm not sure how to record uptime/downtime.

MartinKolarik:
Real-time-only can go into redis in whatever format... Historical data would depend on the selected storage, which likely won't be redis.

jimaek:
So to me it seems for now we need these real-time values:
- CPU load
- Available CPU cores
- In progress tests


Then we need to select a timeseries DB out of the 100 that exist now and start storing:
- Uptime per probe pushed by the API probably
- accepted tests/successful tests/failed tests

[...]

PR #201, opened 2022-08-15
feat: live probe metrics

patrykcieszkowski:
#200 

- [x] probe types
- [x] ws event handler 
- [x] get probes body

relies on https://github.com/jsdelivr/globalping-probe/pull/97

GET /v1/probes
```
[
  {
    "version": "0.9.1",
    "ready": true,
    "location": {
      "continent": "EU",
      "region": "Western Europe",
      "country": "FR",
      "city": "Paris",
      "asn": 12876,
      "latitute": 48.8534,
      "longitude": 2.3488,
      "network": "ONLINE S.A.S."
    },
    "resolvers": [
      "private"
    ],
    "stats": {
      "cpu": {
        "count": 4,
        "load": [
          {
            "usage": 1.02,
            "idle": 98.98
          },
          {
            "usage": 1.98,
            "idle": 98.02
          },
          {
            "usage": 1.03,
            "idle": 98.97
          },
          {
            "usage": 1,
            "idle": 99
          }
        ]
      },
      "jobs": {
        "count": 0
      }
    }
  }
]
```

### DIFF
--- a/src/probe/route/get-probes.ts
+++ b/src/probe/route/get-probes.ts
@@ -27,6 +27,7 @@ const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & R
 			network: socket.data.probe.location.network,
 		},
 		resolvers: socket.data.probe.resolvers,
+		stats: socket.data.probe.stats,
 	}));
 };
 
--- a/src/probe/route/get-probes.ts
+++ b/src/probe/route/get-probes.ts
@@ -10,6 +10,7 @@ type Socket = RemoteSocket<DefaultEventsMap, SocketData>;
 const io = getWsServer();
 
 const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & Router.RouterParamContext>): Promise<void> => {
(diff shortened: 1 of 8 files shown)


src/probe/route/get-probes.ts
```diff
@@ -27,6 +27,7 @@ const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & R
 			network: socket.data.probe.location.network,
 		},
 		resolvers: socket.data.probe.resolvers,
+		stats: socket.data.probe.stats,
```
MartinKolarik, line 30: We agreed with @jimaek to not include this in the public data but instead:
 - [ ] add an optional `adminkey` query param to this endpoint, and include this property only if the `adminkey` is non-empty and correct,
 - [ ] correct value for `adminkey` will be set via config to some string, default is empty, which means the feature isn't available unless some value is explicitly set.

jimaek: And maybe expose the IP address to the admin too. That should simplify our debugging, no more guessing what probe is what

PR #202, opened 2022-08-21
docs: add no probes error

ayuhito:
Are we missing out on any other errors?

### DIFF
--- a/docs/measurement/post-create.md
+++ b/docs/measurement/post-create.md
@@ -92,6 +92,21 @@ headers:
   Location: https://api.globalping.io/v1/measurements/PY5fMsREMmIq45VR
 ```
 
+## error response (no probes found)
+
+**condition**: if provided location doesn't have probes - e.g. `{ locations: [ 'magic': 'not a real place' ] }`
+
+**status code**: `400 Bad Request`
+
+```json
+{
+    "error": {
+      "message": "No suitable probes found",
+      "type": "api_error"
+    }
+}
+```
+
 ## error response (validation failed)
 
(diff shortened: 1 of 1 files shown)


patrykcieszkowski (approved):

patrykcieszkowski:
If we were pedantic, we could document validation errors, but I think they're self explanatory and don't really need that.

```json
{
  "error": {
    "message": "Validation Failed",
    "type": "invalid_request_error",
    "params": {
      "limit": "\"limit\" must be less than or equal to 500"
    }
  }
}
```

MartinKolarik:
Definitely no need to list those but maybe a good idea to document possible `type` values.

ayuhito:
I wonder if no probes should return `"type": "api_error"` in this case. It's the same as 500 Internal Server Error.

At least I was relying on the `type` value to differentiate errors for some GP tools I'm making. But if `type` isn't actually helpful in differentiating all errors, I'd have to rely on status codes instead.

MartinKolarik:
I'm not really sure what the intention with `type` was, but it's true the errors should be divided into categories by status codes, and the `type` would make sense to me if it provided some extra information - either a specific value for each type of error or a more specific category of errors than the status code category. This is something I didn't really review before.

patrykcieszkowski:
> I wonder if no probes should return `"type": "api_error"` in this case. It's the same as 500 Internal Server Error.

that's a valid point. It's not an API error if user queried geo we don't support.

ayuhito:
Let's merge this PR since it works for the current setup and I'll make a new issue to discuss and track what we should do with `type`.

Issue #131, opened 2022-05-22
Better APM


jimaek:
Looks like appsignal is not very helpful, there is lots of data missing, traces are almost empty in most cases. Many errors and issues are not logged at all.
It's too hard to find issues and understand the situation.

The question is, are we doing something wrong or we need to switch to an alternative?

Proposed services to test:
- https://www.datadoghq.com/product/apm/ (includes log integration) +traces every single request
- https://newrelic.com/products/application-monitoring +traces every single request
- https://sentry.io/for/node/
- Other?

patrykcieszkowski:
Artillery is handy for stress testing, but the IP rate-limit might make it more tricky. 
https://www.npmjs.com/package/artillery

jimaek:
I am thinking of using Newrelic due to lower cost and 100% traces at the same time. Plus native logging. Can you explore it @patrykcieszkowski ? 
If we use it we will have to natively integrate to get the fullest picture possible. So no Heroku log drains but logging directly so it could associated logs to traces and so on. Same for full Redis integration and everything else we could use.
Plus our custom charts

patrykcieszkowski:
Newrelic doesn't support ESM as of yet. There are plans to do so, but they have no deadline.

https://github.com/newrelic/node-newrelic/issues/553#issuecomment-1130219375

jimaek:
Since Datadog seems to support everything can you help me make sense of their pricing?
https://dl.dropboxusercontent.com/s/mbzv6dmj8yosmhu/chrome_t1UZegH2gK.png

I think to monitor our API we need:
- APM $36 (but public pricing says $48)
- Infrastructure $18 (required)
- Custom Metrics $5
- Log Management $2.55
- Log Ingestion $1
- Profiled Hosts ?? $23

Total: ~$97 per month?

Am I right or I'm missing something? If this is correct its similar to newrelic and we could look into DataDog too then

jimaek:
Yeah it's a mess and dashboard pricing is not always the same to public...

But if it's under $100/month we can proceed.

Can you check if we can natively integrate with them?
All the fancy stuff like logging with APM correlation and so on. They have a bunch of features and for that kind of money I want to use everything web pay for :p

jimaek:
Another interesting option is https://pm2.io/
They charge per node process and I asked about the enterprise plan. But even the Plus version could be enough in theory.
The only question is how many process we plan to have if we migrate away from Heroku

patrykcieszkowski:
[...]

PR #218, opened 2022-10-31
Migration from AppSignal to New Relic

alexey-yarmosh:
https://github.com/jsdelivr/globalping/issues/131


newrelic.cjs
```diff
@@ -0,0 +1,32 @@
+exports.config = {
```
MartinKolarik, line 1: Don't exclude files from `xo`. If there's a problem with a specific rule, you can always disable it for one line or for the whole file. Here, put `/* eslint-disable camelcase */` at the top.

jimaek:
Do we want to fix the errors here? https://deepsource.io/gh/jsdelivr/globalping/run/236758f2-2948-4ef7-997f-1b4a317fbfdd/javascript/
If not then maybe add those rules to ignore?

PR #245, opened 2022-12-05
fix: validation http codes

### LINKED ISSUE #208: New error type values
Following up on #202, I think having unique `type` values for every error is great for users that depend on the API programmatically. 

Currently, `400 - no probes found` returns `"type": "api_error"` which is the same as the Internal Server Error type. Does `"type": "no_probes_found"` sound suitable?

I also think we should change the `422 - validation error` which returns `"type": "invalid_request_error"` to `"type": "validation_error"` to be more specific.

alexey-yarmosh:
Updated /POST "measurement" validation error to send 400, like on other endpoints.
Updated "No suitable probes found" error code to 422. It is annoying to get 400 both for validation and no probes during usage/dev/test.

Fixes #208

### DIFF
--- a/src/lib/http/middleware/error-handler.ts
+++ b/src/lib/http/middleware/error-handler.ts
@@ -14,7 +14,7 @@ export const errorHandlerMw = async (ctx: Context, next: Next) => {
 			ctx.body = {
 				error: {
 					message: error.expose ? error.message : createHttpError(error.status).message,
-					type: 'api_error',
+					type: error['type'] as string ?? 'api_error',
 				},
 			};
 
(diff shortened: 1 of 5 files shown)


ayuhito:
I'm onboard with the change regarding http codes, but I don't think it addresses #208's original purpose regarding the `type` value.

- If we're going for unique status codes per error, I don't see the value in keeping the `type` value which doesn't allow me to differentiate errors programmatically (since no probes return the same type as a 500 API error). I'll just use status codes.
- Otherwise, it would make sense to have a different `type` value for every unique error we want to disclose. 

Also upon final resolution, please wait on merging this until I update jsdelivr/globalping-chat-bots because some of the error formatting depends on these status codes and types.

alexey-yarmosh:
I update error type messages too, @ayuhito @MartinKolarik if that satisfies the requirements now.

MartinKolarik (approved):

Issue #244, opened 2022-12-01
Review the transport used across API instances


alexey-yarmosh:
Under usual load [@socket.io/redis-adapter](https://github.com/socketio/socket.io-redis-adapter) works fine, however under the high load (>5 RPS * 100 probes) we are able to observe issues like #120 and #234. The root cause is that communication through redis pub/sub becomes very slow under the load. So to address issues above we need to deal with the adapter performance. 

Possible option:
1. tune redis itself
    1.1 run separate redis instances for storage and for API adapter (since their logic does not overlap, we are able to get fast perf boost)
    1.2 use dragonfly as an uderlying db for redis adapter (if the API is compatible it will be quickly to set up, and may give significant perf boost)
2. ~~use another adapter~~ (I tried AMQP and NATS adapter, but they are poorly supported and [don't have required functionality](https://github.com/jsdelivr/globalping/issues/120#issuecomment-1333467970))
3. use another way to communicate with probes (we can try something else instead of socket.io, with the built in horizontal-scaling. [link](https://blog.hotstar.com/building-pubsub-for-50m-concurrent-socket-connections-5506e3c3dabf))
    3.1 NATS
    3.2 MQTT
    3.3 uWebSockets?
    3.4 ...

jimaek:
1. and 1.1: It doesnt seem to be possible to tune it other than clustering to utilize more cores. But that feels like a hack to me because the "high load" we have is not that high really, and if a normal installation cant handle it, adding clustering will only push the problem to the future and not really solve anything
1.2: The perf difference seems minimal I believe
3: Can you expand on that? You want to completely remove socket.io from the system? Or you want to use a queue system for a specific function?

Also uWebSockets is unrelated to "another way of communication", it won't solve any Redis problems.

MartinKolarik:
Clustering to use more cores (if redis can't use all resources it has now) or use independent instances is the correct thing to do as far as redis scaling goes. I agree the current socket setup may not be ideal for our use case but no sure about the alternatives.

jimaek:
I agree about clustering being correct, but I expected to do it in the future to scale to tens of thousands of probes and users, not because it struggles to handle 500 probes and a few requests. So if I scale it now we will only postpone the inevitable by a few months. Based on some online benchmarks Redis is simply super slow at this.

patrykcieszkowski:
[...]

PR #246, opened 2022-12-06
Add 200ms cache to fetchSockets method

alexey-yarmosh:
- fixes https://github.com/jsdelivr/globalping/issues/244
- adds node memory limits according to our current server set up.

### DIFF
--- a/src/lib/ws/server.ts
+++ b/src/lib/ws/server.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import {Server} from 'socket.io';
+import _ from 'lodash';
+import {RemoteSocket, Server} from 'socket.io';
 import {createAdapter} from '@socket.io/redis-adapter';
 import type {DefaultEventsMap} from 'socket.io/dist/typed-events';
 import type {Probe} from '../../probe/types.js';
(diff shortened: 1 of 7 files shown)


src/lib/ws/server.ts
```diff
@@ -42,3 +47,17 @@ export const getWsServer = (): WsServer => {
...
+
+export const fetchSockets = () => {
+	if (!io || !throttledFetchSockets) {
+		throw new Error('WS server not initialized yet');
+	}
+
+	const sockets = throttledFetchSockets();
+
+	if (sockets === undefined) {
+		throw new Error('undefined, the debounced function was not invoked yet');
```
MartinKolarik, line 59: Don't quite see how this would happen. On the first call, it should be invoked right away.

src/lib/ws/server.ts
```diff
@@ -14,8 +15,10 @@ export type WsServer = Server<DefaultEventsMap, DefaultEventsMap, DefaultEventsM
 
 export const PROBES_NAMESPACE = '/probes';
 const TIME_UNTIL_VM_BECOMES_HEALTHY = 8000;
+const TIME_TO_CACHE_FETCH_SOCKETS = 200;
 
 let io: WsServer;
+let throttledFetchSockets: _.DebouncedFunc<() => Promise<Array<RemoteSocket<DefaultEventsMap, SocketData>>>>;
```
MartinKolarik, line 21: You're mixing up words throttle and debounce, which have different meanings (https://css-tricks.com/debouncing-throttling-explained-examples/)

alexey-yarmosh: `_.throttle` itself returns `_.DebouncedFunc`, seems they are using the same type for both methods.
image

src/lib/ws/server.ts
alexey-yarmosh: Me neither. But ts won't let me ignore that.

image

MartinKolarik (approved):

PR #249, opened 2022-12-08
Fix typo in get-probes.ts

HackedServer:
Misspelling of latitude

https://en.wikipedia.org/wiki/Latitude


HackedServer:
Ah it also needs the test updated. Will get to it tomorrow if no one else does first.

jimaek:
Yeah, thanks for sending the PR :) 

@alexey-yarmosh take a look when you can please

PR #250, opened 2022-12-09
System tags

### LINKED ISSUE #236: System tags
It was part of https://github.com/jsdelivr/globalping/issues/153 but now there is an urgent use case that requires this.

Specifically this part:

>System tags. On top of the above we would need to have our own tags that users cant change but can use to query probes. Example: public-dns, regional-dns, datacenter.

The goal at the moment is to tag specific probes that I control with custom tags per probe. I have servers with both AWS and Google Cloud and I need to tag each probe with the "cloud region name", e.g. "eu-west-1", "us-east-2"...

This way a user can then ask to ping from specific cloud regions, e.g. magic:google+eu-west-1

Not really sure how to do it since the system at the moment is ephemeral. It was supposed to be part of the dashboard epic but maybe we can build the system tags first somehow? I already have a mysql DB online for our Slack App. 

But if we do it now i dont want to have to migrate or rewrite anything as we continue on task #153 so it should be thought out. 

@MartinKolarik maybe you have ideas?

alexey-yarmosh:
Resolves #236


jimaek:
Please note that this PR also requires update to /demo to support tags parameter and to all docs that document the new logic

README.md
```diff
@@ -190,16 +187,16 @@ API uses 3000 port by default. This can be overridden by `PORT` environment vari
...
 Once the API is live, you can spin up a probe instance by running the following in the probe repository:
 
 1. Clone [Globalping Probe](https://github.com/jsdelivr/globalping-probe) repository.
 2. `npm install && npm run init:hooks`
-3. `npm run build && NODE_ENV=development node dist/index.js`
+3. Either `npm run build && npm run dev` or `npm run dev:tsx` (no type-checking)
 
 ### Environment Variables
-
+- `PORT=3000` environment variable can be start the API on antoher port (default is 3000)
```
MartinKolarik, line 199:
```suggestion
- `PORT=3000` environment variable can start the API on another port (default is 3000)
```

src/probe/router.ts
```diff
@@ -23,11 +23,23 @@ const locationKeyMap = [
 ];
 
 export class ProbeRouter {
+	static hasIndex(socket: Socket, index: string) {
+		return socket.data.probe.index.some(v => v.includes(index.replace('-', ' ').trim().toLowerCase()));
+	}
+
+	static hasTag(socket: Socket, tag: string) {
+		return socket.data.probe.tags.some(({type, value}) => type === 'system' && value.includes(tag.trim().toLowerCase()));
```
MartinKolarik, line 31: Might be better to normalize the data when building the probe?

alexey-yarmosh: Here `value` is a data from probe it is already normalized. We are only normalizing user input (`tag` variable).

MartinKolarik: True and that seems to be normalized on input as well? At least lowercase, not sure about trimming but that can be moved there as well.

alexey-yarmosh: Removed toLowerCase, trim is necessary here because values from magic comes as string separated by '+' and will not be trimmed by Joi.

MartinKolarik (approved):

PR #257, opened 2022-12-13
Map multiple states to DC iso

### LINKED ISSUE #255: unresolvable geoip
5.161.143.28 (my own server)

```
[2022-12-12 16:06:23] [INFO] [685] [general] Start probe version 0.10.1 in a production mode
[2022-12-12 16:06:23] [ERROR] [685] [general] connection to API failed: unresolvable geoip
```

Why unresolvable? Isn't our logic to fallback to ipinfo? In which case it would be marked as Washington https://ipinfo.io/5.161.143.28
Digitalelement resolves it as well, only maxmind doesnt. So why in case of 2 successful resolutions we fail?

alexey-yarmosh:
Fixes #255


src/lib/location/location.ts
```diff
@@ -32,17 +32,17 @@ export const getContinentByCountry = (country: string): string => {
...
+		throw new Error(`iso not found ${state}`);
 	}
 
-	return String(stateEntries[0]);
+	return iso;
 };
 
 export const getStateNameByIso = (iso: string): string => {
-	const state = states[iso as keyof typeof states];
+	const state = _.invert(states)[iso];
```
MartinKolarik, line 45: Hmm, so we're using this the other way too. Ok, I guess but should add a test to check that DC resolves to "District Of Columbia" otherwise, we could accidentally change it later just be reordering the lines.

PR #258, opened 2022-12-13
docs: match error codes with current api

ayuhito:
Follow up PR to #245 to sync up docs with the current API.


MartinKolarik:
Already covered in #259

PR #261, opened 2022-12-16
fix: ignore all updates after measurement finished

### LINKED ISSUE #254: If measurement finishes by timeout its counter is not cleared
Additionally no TTL is set so these records just eat space https://dl.dropboxusercontent.com/s/q3kvn3yhotj61m9/RedisInsight-v2_2022-12-10_19-02-11.png

alexey-yarmosh:
Fixes #254

### DIFF
--- a/src/measurement/runner.ts
+++ b/src/measurement/runner.ts
@@ -63,19 +63,21 @@ export class MeasurementRunner {
 	}
 
 	async recordResult(data: MeasurementResultMessage): Promise<void> {
-		await this.store.storeMeasurementResult(data);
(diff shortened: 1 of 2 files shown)


MartinKolarik (approved):

jimaek:
We should also drop all previous stuck records, any recommendations?

alexey-yarmosh:
We can simply FLUSHDB, or write a script to delete all non-positive `probes_awaiting`

jimaek:
Flushdb is too destructive, let's target only the broken records

PR #271, opened 2023-01-04
docs: remove trailing "/" in urls

DavidEredics:
Some urls in the docs have a "/" at the end when it shouldn't and therefore it doesn't work.


alexey-yarmosh (approved):
Thanks!

PR #272, opened 2023-01-04
docs: fix link in docs to the right file

DavidEredics:
In the docs there is a link to a wrong filename. Edited to point to the right file.


alexey-yarmosh (approved):
Thanks!

Issue #212, opened 2022-10-05
Empty results of some probes


alexey-yarmosh:
Sometimes some of the probes of the `finished` measurements have result fields. Maybe data is not sent by the probe or API didn't handle it properly. We need to adress it.

Steps to reproduce (may not reproduce every time):
- send measurement request with the body
```json
{
    "target": "yarmosh.by",
    "type": "ping",
    "measurementOptions": {
        "packets": 16
    },
    "limit": 500,
    "locations": []
}
```
- wait until measurement `status` is `finished`
- check all the `result` fields
Expected value:
some valid result data
Actual value:
<img width="1019" alt="image" src="https://user-images.githubusercontent.com/16905445/194159450-9fed85da-a704-48cf-bd92-339e075a2d05.png">

jimaek:
The question is why it timed out. Our current probes should be reliable enough to work. If you give me some probes I can try to pull the local logs.

We probably need a status field to clearly mark what happened in each result. Did the whole probe time out because of bad internet, meaning the target endpoint is not at fault, did it time out because of an exception in which case we need to fix it, did the API fail and we need to fix something, did the target endpoint fail and the user can safely mark it as down

patrykcieszkowski:
> Isn't this a case of the measuremnt timing out? I believe we don't have any explicit way to indicate per-measurement errors so this is what happens then.

no. A measurement timeout is handled by the server, and the server doesn't include any other value than `rawOutput`.

alexey-yarmosh:
Such output may occur for several reasons:
- stdout of the result is empty [(code)](https://github.com/jsdelivr/globalping-probe/blob/master/src/command/ping-command.ts#L110)
- stdout of the execa error is empty [(code)](https://github.com/jsdelivr/globalping-probe/blob/master/src/command/ping-command.ts#L117)
- error happened but it is not execa error [(code)](https://github.com/jsdelivr/globalping-probe/blob/master/src/command/ping-command.ts#L117)

Also we are putting the reason of failed measurement (e.g. "Private IP ranges are not allowed") in `rawOutput` which is used for execa stdout.

I think we should add a `status` field to the measurement + use `rawOutput` only for execa output and also add `message` field for the reason of the failure. E.g.:
- successful measurement
```
"result": {
  "status": "success",
  "message": "",
  "rawOutput": "PING ...",
  "resolvedAddress": "...",
  "resolvedHostname": "...",
  "timings": [...],
  "stats": {...}
}
```
[...]

PR #283, opened 2023-02-01
feat: add measurement statuses

alexey-yarmosh:
Adds status to the result field of the probe measurement. Possible values are: in-progress, finished, and failed.

Probe part is here: https://github.com/jsdelivr/globalping-probe/pull/108

Part of #212

### DIFF
--- a/src/measurement/types.ts
+++ b/src/measurement/types.ts
@@ -7,7 +7,7 @@ import type {Location} from '../lib/location/types.js';
 
 type TestResult = {
 	rawOutput: string;
-	status: 'in-progress' | 'success' | 'failed';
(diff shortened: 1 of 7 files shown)


src/measurement/types.ts
```diff
@@ -7,6 +7,7 @@ import type {Location} from '../lib/location/types.js';
 
 type TestResult = {
 	rawOutput: string;
+	status: 'in-progress' | 'success' | 'failed';
```
MartinKolarik, line 10: "success" should be "finished"

jimaek:
Please also review the docs in case anything needs to be updated

PR #293, opened 2023-02-27
feat: scan for timed out measurements instead of storing in memory

### LINKED ISSUE #229: Review timeout logic
https://github.com/jsdelivr/globalping/issues/176#issuecomment-1199235237

>It spawns an internal timer per every measurement, and holds them locally. If the API goes down, the timer is gone, and the status is never marked as finished, unless the result comes - which is not guaranteed.

>A way better solution would be to have a global interval running per each API instance, and just query unresolved measurements older than X.

### LINKED ISSUE #292: Invalid JSON content of example of HTTP type measurement
In the example of **HTTP** type, there is missing nesting of  request parameters into **measurementOptions** node.
A correct example might look like this

```
{
  "type": "http",
  "target": "jsdelivr.com",
  "locations": [
    {
      "continent": "eu",
      "limit": 3
    }
  ],
  "measurementOptions": {
    "request": {
      "path": "/",
      "query": "?a=abc",
      "method": "GET",
      "host": "jsdelivr.com",
      "headers": {
        "Referer": "https://example.com/"
      }
    },
    "port": 443,
    "protocol": "HTTPS"
  }
}
```

alexey-yarmosh:
Fixes #229 
Additionally fixes #292


src/measurement/store.ts
```diff
@@ -99,24 +104,62 @@ export class MeasurementStore {
...
+
+			if (cursor !== 0) {
+				logger.warn(`There are more than ${SCAN_BATCH_SIZE} "in-progress" elements in db`);
+			}
+
+			const timedOutIds = tuples
+				.filter(({value: time}) => Date.now() - Number(time) >= timeoutTime)
+				.map(({field: id}) => id);
+
+			await this.markFinishedByTimeout(timedOutIds);
```
MartinKolarik, line 160: We need to catch and log+ignore errors here. Otherwise, one error anywhere kills the process.

PR #300, opened 2023-03-21
feat: increase ram availiable to node process

alexey-yarmosh:


jimaek:
Closing as it won't be needed due to multi threaded API

PR #323, opened 2023-04-06
docs: add client guidelines

MartinKolarik:


jimaek:
LGTM

We should in general rewrite the docs later on. They're hard to read

alexey-yarmosh (approved):

MartinKolarik:
> We should in general rewrite the docs later on. They're hard to read

Will. But for now at least collecting the important points.

Issue #247, opened 2022-12-06
Smarter magic field


jimaek:
The magic field should get a bit more logic to make the output more predictable. e.g. if you do magic:DE you will match not just Germany but every city and ISP that has "de" in their names and will get a random output.

At first I thought of a complicated system with weights, but after thinking about it I think there's a better one.

It sounds like the simplest solution would be to first sort the data in a way that makes most sense and then allow the system to select the first N matching values. e.g. a version of sorting could be:

1. ISO country names, DE
2. Country names, Germany
2. Cities, Berlin
3. US state, Texas
4. Continent, Europe
5. System tags
6. ISP names, Netia
7. ASN, 47875

This way if you do magic: DE limit:10 you should in theory just get 10 probes with ISO=DE and the check would never even get to "Belgrade" and "Clouvider" which also contain the text "DE". But if you do magic:DE with limit 1000 then its your problem that you will get random results and the solution is to use strict parameters and not magic.

BUT the data within each category should be random to avoid things where a single probe gets 100% of tests that ask for 1 measurement. 

I think the task itself should also include automated tests to avoid regression.

1. Write automated CI tests that ensure different combos of filters work correctly
8. Exclude user tags from magic
9. Add ranking system?

patrykcieszkowski:
You could also add a support for optional type matching. `type=value`.

Then you don't have to sort or assume anything, and your results are more reliable.

jimaek:
If the person knows what they're looking for they should just use the static parameters anyway. This is supposed to be a human friendly input where you can run something that you think should work and it works. Adding special formatting is probably not the best, it should just magically work :)

alexey-yarmosh:
Looks valid, we can also combine that with the priority for full string match. To avoid bugs when city (or another lower priority field) fully sits inside the country name.

alexey-yarmosh:
>As per value sorting, here's their current order:
>https://github.com/jsdelivr/globalping/blob/master/src/probe/builder.ts#L76-L89
>I think most of them are in their correct order already.

Does that affect the order? As I see we are only filtering by the values from index.
https://github.com/jsdelivr/globalping/blob/master/src/probe/router.ts#L60

patrykcieszkowski:
> Does that affect the order? As I see we are only filtering by the values from index. https://github.com/jsdelivr/globalping/blob/master/src/probe/router.ts#L60
[...]

Issue #334, opened 2023-04-12
Setting limit to higher than available probes return data from other regions


perosb:
Setting the limit to f.ex 3 and only want to test from Vietnam that only have 2 probes, returns result from a third arbitrary probe.
Maybe this should either:
1. throw error
2. return result from one of the existing probes in VN again (might be troublesome if limit 1000 hammering probe)

I'm thinking this might give incorrect analysis unless you're observant.

As a comparison, selecting a country with 0 probes shows an error.

```
~ $ globalping ping jsdelivr.com from VN --limit 3
> AS, VN, Muong Lay, ASN:45899, VNPT Corp
PING jsdelivr.com (104.21.35.47) 56(84) bytes of data.
64 bytes from 104.21.35.47 (104.21.35.47): icmp_seq=1 ttl=55 time=29.0 ms

> AS, VN, Hanoi, ASN:135905, VIETNAM POSTS AND TELECOMMUNICATIONS GROUP
PING jsdelivr.com (104.21.35.47) 56(84) bytes of data.
64 bytes from 104.21.35.47 (104.21.35.47): icmp_seq=1 ttl=58 time=19.1 ms

> EU, HU, Budapest, ASN:30836, 23VNet Kft.
PING jsdelivr.com (104.21.35.47) 56(84) bytes of data.
64 bytes from 104.21.35.47 (104.21.35.47): icmp_seq=1 ttl=58 time=20.7 ms
```

edit: just noticed that setting `from #ASN` with high limit only return the correct ASN

jimaek:
Its not arbitrary, it returns the last one because "23VNet" contains "VN" and it matches it

jimaek:
I think you could just use Vietnam as input and it would fix the issue :)

perosb:
Ah, that explains this scenario but might be more confusing.

None the less, this was just a suggestion for improvements.

MartinKolarik:
This is confusing and Vietnam might break as well in the future, which is why I said partial matches should be ignored if there's an exact match. E.g., if the input is an ISO code, nothing else than ISO codes should be considered:  https://github.com/jsdelivr/globalping/issues/247#issuecomment-1339815722 We need to revisit that.

patrykcieszkowski:
> but I cant imagine how a user would use them

```globalping ping jsdelivr.com from country=VN --limit 3```

jimaek:
> > but I cant imagine how a user would use them
> 
[...]

PR #340, opened 2023-04-18
feat: ignore partial matches after exact match in magic field

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/247 and https://github.com/jsdelivr/globalping/issues/334

### DIFF
--- a/test/tests/unit/probe/router.test.ts
+++ b/test/tests/unit/probe/router.test.ts
@@ -466,7 +466,7 @@ describe('probe router', () => {
 			expect(probes[2]!.location.country).to.equal('CZ');
 		});
 
-		it('should ignore matches on the lower levels after exact match', async () => {
+		it('should ignore low-priority partial matches if there is an exact match', async () => {
 			const sockets: Array<DeepPartial<Socket>> = [
 				await buildSocket('socket-3', { country: 'VN', normalizedCity: 'hanoi', normalizedNetwork: 'ultra networks' }),
 				await buildSocket('socket-2', { country: 'RU', normalizedCity: 'vnukovo', normalizedNetwork: 'super networks' }),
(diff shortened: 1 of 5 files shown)


test/tests/unit/probe/router.test.ts
```diff
@@ -447,37 +447,73 @@ describe('probe router', () => {
...
+			fetchSocketsMock.resolves(sockets as never);
+
+			const probes = await router.findMatchingProbes([
+				{ magic: 'vn' },
+			], 100);
+
+			expect(fetchSocketsMock.calledOnce).to.be.true;
+			expect(probes.length).to.equal(1);
+			expect(probes[0]!.location.country).to.equal('VN');
+		});
```
MartinKolarik, line 484: This matches what I described in my last comment, but now I'm thinking it would make more sense even here to include only the exact match (https://github.com/jsdelivr/globalping/issues/247#issuecomment-1339815722). The impact isn't big either way, though, as this is mostly likely going to happen only with networks.

test/tests/unit/probe/router.test.ts
```diff
@@ -447,37 +447,73 @@ describe('probe router', () => {
...
+			const probes = await router.findMatchingProbes([
+				{ magic: 'wars' },
+			], 100);
+
+			expect(fetchSocketsMock.calledOnce).to.be.true;
+			expect(probes.length).to.equal(3);
+			expect(probes[0]!.location.country).to.equal('PL');
+			expect(probes[1]!.location.country).to.equal('PL');
 			expect(probes[2]!.location.country).to.equal('PL');
 		});
```
MartinKolarik, line 503: Sorry highlighted the wrong test before:

This matches what I described in my last comment, but now I'm thinking it would make more sense even here to include only the exact match (https://github.com/jsdelivr/globalping/issues/247#issuecomment-1339815722). The impact isn't big either way, though, as this is most likely going to happen only with networks.

MartinKolarik: Note that a more interesting scenario with this problem may be:

```js
const sockets: Array<DeepPartial<Socket>> = [
	await buildSocket('socket-1', { country: 'PL', normalizedCity: 'warsaw', normalizedNetwork: 'super networks' }),
	await buildSocket('socket-2', { country: 'PL', normalizedCity: 'warsaw', normalizedNetwork: 'super networks' }),
	await buildSocket('socket-3', { country: 'PL', normalizedCity: 'poznan', normalizedNetwork: 'wars' }),
];
```

With the current version of this PR and `limit: 1`, it would always return one of the Warsaw probes because cities have higher priority than networks. If we change it to "exact match always wins", it would return Poznan in this case.

MartinKolarik: More realistic examples that still don't work as expected are cities "Oi" vs "Hanoi" and "New York City" vs "York".

alexey-yarmosh: We can either do "if there's an exact match, ignore partial matches" or "give the exact matches highest priority over index fields priorities".

MartinKolarik: Let's go with first, regardless of sorting. Sorting will be used only if there's no exact match.

alexey-yarmosh: done

MartinKolarik (approved):

PR #349, opened 2023-04-30
Add OpenAPI setup

MartinKolarik:
Prepares a setup for documenting the API using an OpenAPI document based on https://github.com/jsdelivr/data.jsdelivr.com/pull/55:
 - [spectral](https://github.com/stoplightio/spectral) is used for validating the OpenAPI document,
 - [portman](https://github.com/apideck-libraries/portman/) is used to run basic tests over HTTP to check the documented operations are actually implemented,
 - an additional chai plugin is used to validate schemas for all responses in the integration tests; this ensures all the various response types match the documentation.

Note that the chai plugin and the spectral config are just copied over from https://github.com/jsdelivr/data.jsdelivr.com for now. After finishing the docs (in a separate PR), I'll maybe move them to a module installed via npm.


test/tests/integration/probes/get-probes.test.ts
```diff
@@ -46,6 +46,7 @@ describe('Get Probes', () => {
 				.expect(200)
 				.expect((response) => {
 					expect(response.body).to.deep.equal([]);
+					expect(response).to.matchApiSchema();
```
MartinKolarik, line 49: Only used here for now to test the integration. Once the other endpoints are documented, will be used in all tests.

alexey-yarmosh (approved):

Issue #348, opened 2023-04-28
Add fields to measurement responses


MartinKolarik:
To support sharing of the measurement results and easily re-running the measurements later, we need to store the actual configuration that was used for creating the measurement and include it in the responses. Store the following fields that are part of the `POST /measurements` body and return them in `GET /measurements/:id` responses:

 - target,
 - limit,
 - locations,
 - measurementOptions.

Each field should only have non-values. Otherwise, it should be omitted.

PR #350, opened 2023-05-02
feat: add measurement request data to the response object

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/348

### DIFF
--- a/src/measurement/store.ts
+++ b/src/measurement/store.ts
@@ -6,6 +6,7 @@ import type { RedisClient } from '../lib/redis/client.js';
 import { getRedisClient } from '../lib/redis/client.js';
 import { scopedLogger } from '../lib/logger.js';
 import type { MeasurementRecord, MeasurementResultMessage, MeasurementResult, MeasurementRequest } from './types.js';
+import { getDefaults } from './schema/utils.js';
 
 const logger = scopedLogger('store');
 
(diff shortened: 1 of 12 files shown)


src/measurement/store.ts
```diff
@@ -38,11 +39,15 @@ export class MeasurementStore {
 			this.redis.set(getMeasurementKey(id, 'probes_awaiting'), probes.length, { EX: probesAwaitingTtl }),
 			this.redis.json.set(key, '$', {
 				id,
-				type,
+				type: request.type,
 				status: 'in-progress',
 				createdAt: startTime.toISOString(),
 				updatedAt: startTime.toISOString(),
+				target: request.target,
+				...(request.locations.some(l => l.limit) ? {} : { limit: request.limit }),
```
MartinKolarik, line 47: In either case, the `limit` should be included only if non-default.

src/measurement/store.ts
```diff
@@ -38,11 +39,15 @@ export class MeasurementStore {
...
 				probesCount: probes.length,
+				...(request.locations.length ? { locations: request.locations } : {}),
```
MartinKolarik, line 49: In either case, the limit should be included only if non-default.

src/measurement/store.ts
```diff
@@ -38,11 +39,15 @@ export class MeasurementStore {
...
+				...(_.isEmpty(request.measurementOptions) ? {} : { measurementOptions: request.measurementOptions }),
```
MartinKolarik, line 50: Looks like this doesn't work either because the request includes fields added by Joi at this point. We'll again have to check for non-default values.

Issue #372, opened 2023-06-09
nock package should throw if not mocked


alexey-yarmosh:
Currently if nock request is not mocked it will do a real request to the 3rd party API. Instead it should throw an error. Need to reconfigure it.

PR #373, opened 2023-06-12
fix: fix ip geo api error handling

alexey-yarmosh:

### DIFF
--- a/src/lib/geoip/client.ts
+++ b/src/lib/geoip/client.ts
@@ -63,7 +63,7 @@ export default class GeoipClient {
 
 		const resultsWithCities = results.filter(s => s.city);
 
-		if (resultsWithCities.length < 2 && resultsWithCities[0]?.provider === 'fastly') {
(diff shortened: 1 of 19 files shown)


src/lib/geoip/client.ts
```diff
@@ -63,7 +63,7 @@ export default class GeoipClient {
 
 		const resultsWithCities = results.filter(s => s.city);
 
-		if (resultsWithCities.length < 2 && resultsWithCities[0]?.provider === 'fastly') {
+		if (resultsWithCities.length === 0 || (resultsWithCities.length === 1 && resultsWithCities[0]?.provider === 'fastly')) {
```
alexey-yarmosh, line 66: Without that update if `resultsWithCities` is empty (all 3 geo providers failed) we will not throw an error and fail further with
```
[2023-06-12 13:57:49] [ERROR] [15402] [geoip] failed to find a correct value for a field "normalizedCity"
```
instead, we want to throw and see an explicit error
```
[2023-06-12 13:58:53] [ERROR] [15554] [probe-metadata] unresolvable geoip: 95.155.94.127
{ expose: true }
Error: unresolvable geoip: 95.155.94.127
```

PR #375, opened 2023-06-12
 fix: fix ip geo api error handling

alexey-yarmosh:
V2 PR to trigger CI build.

### DIFF
--- a/config/test.cjs
+++ b/config/test.cjs
@@ -7,6 +7,11 @@ module.exports = {
 	admin: {
 		key: 'admin',
 	},
+	geoip: {
+		cache: {
+			ttl: 1, // 1 ms ttl here to disable redis cache in tests
+		},
+	},
 	ws: {
 		fetchSocketsCacheTTL: 0,
 	},
--- a/src/lib/geoip/client.ts
+++ b/src/lib/geoip/client.ts
@@ -63,7 +63,7 @@ export default class GeoipClient {
 
 		const resultsWithCities = results.filter(s => s.city);
 
-		if (resultsWithCities.length < 2 && resultsWithCities[0]?.provider === 'fastly') {
+		if (resultsWithCities.length === 0 || (resultsWithCities.length === 1 && resultsWithCities[0]?.provider === 'fastly')) {
 			throw new InternalError(`unresolvable geoip: ${addr}`, true);
 		}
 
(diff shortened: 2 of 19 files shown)


src/lib/geoip/client.ts
```diff
@@ -63,7 +63,7 @@ export default class GeoipClient {
 
 		const resultsWithCities = results.filter(s => s.city);
 
-		if (resultsWithCities.length < 2 && resultsWithCities[0]?.provider === 'fastly') {
+		if (resultsWithCities.length === 0 || (resultsWithCities.length === 1 && resultsWithCities[0]?.provider === 'fastly')) {
```
alexey-yarmosh, line 66: Without that update if resultsWithCities is empty (all 3 geo providers failed) we will not throw an error and fail further with
```
[2023-06-12 13:57:49] [ERROR] [15402] [geoip] failed to find a correct value for a field "normalizedCity"
```
instead, we want to throw and see an explicit error
```
[2023-06-12 13:58:53] [ERROR] [15554] [probe-metadata] unresolvable geoip: 95.155.94.127
{ expose: true }
Error: unresolvable geoip: 95.155.94.127
```

config/test.cjs
```diff
@@ -7,6 +7,11 @@ module.exports = {
 	admin: {
 		key: 'admin',
 	},
+	geoip: {
+		cache: {
+			ttl: 1, // 1 ms ttl here to disable redis cache in tests
```
MartinKolarik, line 12: > Additionally redis cached some of the geo ip values during tests, which made them pass locally and fail on CI.

Do you mean from previous tests or from dev runs? Because this will fix test mixups, but if something is already cached in redis, it'll stay there (unless we delete it which I don't think we do). In other projects, I solve this by calling flushDb() in dev/test envs on startup.

alexey-yarmosh: Both, `ttl: 1` is to fix previous test mixups. And [fakeIpForDebug](https://github.com/jsdelivr/globalping/pull/375/files#diff-24fca2bd712e62250531a7ec03d88f45b2874994059c608fda0712b1f534f87eR25) update is to fix prev dev runs cache. flushDb seems to be more elegant, so I will replace the second update with it.

MartinKolarik: 👍 here for reference: https://github.com/jsdelivr/data.jsdelivr.com/blob/master/src/lib/redis/index.js#L10-L12

PR #376, opened 2023-06-12
feat: fix ip geo api error handling

alexey-yarmosh:
V2 PR to trigger CI build.


alexey-yarmosh:
This is a duplicate of https://github.com/jsdelivr/globalping/pull/375

Issue #264, opened 2022-12-19
IP Geo tools to verify progress


jimaek:
We have GEO detection problems but we also dont have a simple way to compare and benchmark different DBs and algorithms.

~~I believe @MartinKolarik wrote some scripts to simplify testing, maybe we can build on top of them to finally get a tool that can reliably tell us if a change resulted in improved data or not.~~

e.g. stuff we could test would be using more DBs, like going all the way to 5. Plus a sixth ipmap data https://github.com/jsdelivr/globalping/issues/248
Later a 7th source of data a traceroute API and so on.

Example wrong IP which is really in Krakow:

Poznan https://ipinfo.io/37.47.67.95
Krakow https://www.maxmind.com/en/geoip2-precision-demo?ip_address=37.47.67.95
Dabrowa https://globalping-geoip.global.ssl.fastly.net/37.47.67.95
Tarnow https://db-ip.com/37.47.67.95
Krakow https://www.ip2location.com/demo/37.47.67.95

MartinKolarik:
The script I wrote simply called the three resolvers we have here and logged when some gave different answers than others. As discussed, we'll start here by creating a list of IPs with verified locations.

alexey-yarmosh:
If I understand correctly we need a script that will check different ip geo providers and accumulation algorithms against the 100% valid data we have. So as a result we will get a data and a script to generate the data again.

Format:
| ip        | real city | ipinfo    | maxmind   | fastly | provider#4 | algorithm#1 | algorithm#2 |
|-----------|-----------|-----------|-----------|--------|------------|-------------|-------------|
| 1.1.1.1   | Warsaw    | Warsaw    | Krakow    | Krakow | Warsaw     | Warsaw      | Krakow      |
| 2.2.2.2   | Barselona | Barselona | Barselona | Madrid | Madrid     | Barselona   | Madrid      |
| ...       | ...       | ...       | ...       | ...    | ...        | ...         | ...         |
| Accuracy: | 1         | 1         | 0.5       | 0      | 0.5        | 1           | 0           |

jimaek:
That would be a good start yes. The only problem is that the verified DB is too small for now

alexey-yarmosh:
Additionally we will see the amount of probes which have different locations from different providers, and understand the scale of a problem.

jimaek:
yes, we can feed all connected probes into the script to see what we're dealing with

alexey-yarmosh:
[...]

PR #382, opened 2023-06-28
feat: add ip2location and ipmap geo ip providers

alexey-yarmosh:
Part of #264

Note:
`IP_2_LOCATION_API_KEY=value` env var should be provided

Priority of providers in desc order:
`[ 'ip2location', 'ipmap', 'maxmind', 'ipinfo', 'fastly' ]`

Logic:
- providers data is grouped by `normalizedCity` value
- provider with the highest priority in a biggest group wins
- if there is no network data it is fulfilled by other provider with the same `normalizedCity`
- if there is no any network data probe is unresolvable
- if there is only city data from `fastly` probe is unresolvable


MartinKolarik:
Please check the conflicts.

alexey-yarmosh:
Fixed

src/lib/geoip/client.ts
```diff
@@ -126,40 +133,53 @@ export default class GeoipClient {
...
-		// Group by the same field value
-		const grouped = Object.values(_.groupBy(filtered, field));
-		// Move items with the same values to the beginning
-		const ranked = grouped.sort((a, b) => b.length - a.length).flat();
+		// Initially sort sources so they are sorted inside groups
+		const sorted = filtered.sort((sourceA, sourceB) => {
+			const indexSourceA = DESC_PRIORITY_OF_PROVIDERS.indexOf(sourceA.provider);
+			const indexSourceB = DESC_PRIORITY_OF_PROVIDERS.indexOf(sourceB.provider);
+			return indexSourceA - indexSourceB;
+		});
```
MartinKolarik, line 157: The priority is already established by the order in https://github.com/jsdelivr/globalping/pull/382/files#diff-026ba18a99076807dbf631c38a550192cb83dcbeb30cbb3521f9539bb8eedac9R56-R62 so I don't think this is necessary.

src/lib/geoip/client.ts
```diff
@@ -126,40 +133,53 @@ export default class GeoipClient {
...
+				return smallestIndexSourceA - smallestIndexSourceB;
+			}
 
-		// If all values are different
-		if (grouped.length === filtered.length) {
-			const sourcesObject = Object.fromEntries(filtered.map(s => [ s.provider, s ]));
-			best = sourcesObject['ipinfo'] ?? sourcesObject['maxmind'];
-		}
+			return groupsSizeDiff;
+		}).flat();
```
MartinKolarik, line 172: Same here. It seems you just wrote precisely what `const ranked = grouped.sort((a, b) => b.length - a.length).flat();` did already, but in a more complicated way.

src/lib/geoip/client.ts
```diff
@@ -126,40 +133,53 @@ export default class GeoipClient {
...
-				network: maxmind.network,
-				normalizedNetwork: maxmind.normalizedNetwork,
-			};
+		for (const source of rankedSources) {
+			if (source.normalizedCity === best.normalizedCity && source?.asn && source?.network) {
+				return {
+					asn: source.asn,
+					network: source.network,
+					normalizedNetwork: source.normalizedNetwork,
+				};
```
MartinKolarik, line 142: Not sure if this was discussed anywhere, but the previous logic intentionally always used ASN data from MaxMind because we considered it the best.

src/lib/geoip/client.ts
```diff
@@ -41,16 +44,20 @@ export default class GeoipClient {
...
-				this.lookupWithCache<LocationInfo>(`geoip:ipinfo:${addr}`, async () => ipinfoLookup(addr)),
+				this.lookupWithCache<LocationInfo>(`geoip:ip2location:${addr}`, async () => ip2LocationLookup(addr)),
+				this.lookupWithCache<LocationInfo>(`geoip:ipmap:${addr}`, async () => ipmapLookup(addr)),
 				this.lookupWithCache<LocationInfo>(`geoip:maxmind:${addr}`, async () => maxmindLookup(addr)),
+				this.lookupWithCache<LocationInfo>(`geoip:ipinfo:${addr}`, async () => ipinfoLookup(addr)),
 				this.lookupWithCache<FastlyBundledResponse>(`geoip:fastly:${addr}`, async () => fastlyLookup(addr)),
 			])
-			.then(([ ipinfo, maxmind, fastly ]) => {
-				const fulfilled = [];
+			.then(([ ip2location, ipmap, maxmind, ipinfo, fastly ]) => {
```
MartinKolarik, line 53: I see the order changes here and shifts ipinfo from the first position to the fourth. I understand it's based on the Google table, but with the radius matching (not yet implemented here), it seems ipinfo is again the best one - do you plan to then reorder this again?

alexey-yarmosh: It depends on do we want to vote before the radius matching (using the original provider cities) or after (using the radius approximated cities). I was thinking about the first option [initialy](https://github.com/jsdelivr/globalping/issues/264#issuecomment-1604376222), but second makes sense too.

Nothing prevents us from implementing the approximation first and then take a look at the results. Reordering of steps as well as reprioritising of providers can be done easily.

src/lib/geoip/client.ts
MartinKolarik: I'm not saying this is wrong but the previous way was on purpose. We considered ASN data from other providers so bad that we decided it was better to not resolve the probe than to use them. @jimaek wdyt?

src/lib/geoip/client.ts
MartinKolarik: I asked because based on all our previous observations, ipinfo seemed to be the most precise one. The Google table currently shows something else but it still isn't a very big sample and ipinfo jumps back to the top again if we consider radiuses - so @jimaek and I still think it should have the highest priority.

alexey-yarmosh: Best radius (60 km) was chosen based on the ipinfo data. Then that radius was applied to other providers, so approximation results may be a bit bound to ipinfo. But anyway if we are comparing by approximated cities priorities should be changed and ipinfo is the most accurate right now.

src/lib/geoip/client.ts
jimaek: I guess the ASN data from the winner does make sense, but I would push ipinfo to a much higher priority. Its the only one that is actually under active development and keeps adding improvements to their API and data.

alexey-yarmosh: I can switch the priority along with the radius approximation logic, as ipinfo shows great results exactly when radius logic is applied.

alexey-yarmosh:
@MartinKolarik as I see PR is ready for merge and deploy.

Issue #388, opened 2023-07-07
Probes dont reconnect after API returns 503 errors


jimaek:
Basically the API went down and no probes tried to reconnect after the API was fixed

```
2023-07-07 11:11:40] [DEBUG] [682] [general] ping request Gy5AIIUvdzt6Gj1C received
[2023-07-07 11:12:05] [DEBUG] [682] [general] ping request 2xYkrLtrnrbIH8ez received
[2023-07-07 11:12:57] [DEBUG] [682] [api:error] disconnected from API: transport close
[2023-07-07 11:12:57] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:12:58] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:12:58] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:12:58] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:12:59] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:12:59] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:00] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:00] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:01] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:02] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:02] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:03] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:03] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:04] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:04] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:05] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[2023-07-07 11:13:05] [ERROR] [682] [general] connection to API failed: Unexpected server response: 503
[...]

PR #389, opened 2023-07-10
fix: make geoip client work even if redis is broken

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/388

At the time of the issue redis was in a strange state, it was rejecting all the queries (GET, SET, etc.). Our geo ip logic uses redis to get cached value before making real requests to the 3rd party geo APIs. Since get command was throwing errors the whole geo ip function was throwing and API were unable to get the geo ips of probes and sent `unresolvable geoip` to them.

Probes were constantly reconnecting when API was down. But they stopped reconnecting after getting `unresolvable geoip` error from the API, which is valid behaviour from the perspective of the probe.

In that fix we are catching any geo ip redis errors and making request to the 3rd party geo APIs not only if there is no value in the cache but also if the cache itself is buggy/unavailable.

### DIFF
--- a/src/lib/geoip/client.ts
+++ b/src/lib/geoip/client.ts
@@ -1,6 +1,5 @@
 import _ from 'lodash';
 import config from 'config';
-import type { Logger } from 'winston';
 import newrelic from 'newrelic';
 import type { CacheInterface } from '../cache/cache-interface.js';
 import { InternalError } from '../internal-error.js';
@@ -27,16 +26,12 @@ export type NetworkInfo = {
 	asn: number;
 };
 
-export const createGeoipClient = (): GeoipClient => new GeoipClient(
-	new RedisCache(getRedisClient()),
-	scopedLogger('geoip'),
-);
+const logger = scopedLogger('geoip');
+
+export const createGeoipClient = (): GeoipClient => new GeoipClient(new RedisCache(getRedisClient()));
 
 export default class GeoipClient {
-	constructor (
-		private readonly cache: CacheInterface,
-		private readonly logger: Logger,
-	) {}
+	constructor (private readonly cache: CacheInterface) {}
 
 	async lookup (addr: string): Promise<ProbeLocation> {
 		const results = await Promise
@@ -155,7 +150,7 @@ export default class GeoipClient {
 		}
 
 		if (!best || best.provider === 'fastly') {
-			this.logger.error(`failed to find a correct value for a field "${field}"`, { field, sources });
+			logger.error(`failed to find a correct value for a field "${field}"`, { field, sources });
 			throw new Error(`failed to find a correct value for a field "${field}"`);
 		}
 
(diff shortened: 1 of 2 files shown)


MartinKolarik:
> But they stopped reconnecting after getting unresolvable geoip error from the API, which is valid behaviour from the perspective of the probe.

We should try to reconnect with some longer delay here as well to handle occasional 3rd party API failures, geo data updates, etc. If the probes don't attempt to reconnect at all, I'd suggest adding a retry after 1 hour or so for all such cases.

MartinKolarik (approved):

alexey-yarmosh:
Seems like the "stop reconnecting in case of unresolvable geoip" logic is not intentional. [Here](https://github.com/jsdelivr/globalping-probe/blob/master/src/index.ts#L83-L89) is the explicit code to stop reconnects in some cases, but `unresolvable geoip` is not in the `fatalConnectErrors` list. Real reason why it is not reconnecting is because [socketio not reconnecting for errors from middlewares](https://socket.io/docs/v4/client-socket-instance/#connect_error). It may also not reconnect for other similar reason. So we need to add an explicit `socket.connect()` to fix that:
```js
if (isFatalError) {
  // At that stage socket.connected=false already,
  // but we want to stop reconnections for fatal errors
  socket.disconnect();
} else {
  socket.connect();
}
```

If we want to persist the current not reconnect logic, we can add `unresolvable geoip` to `fatalConnectErrors` list. But such probes are reconnected only once a week, after a restart.

MartinKolarik:
The error is changed to `failed to collect probe metadata` at https://github.com/jsdelivr/globalping/blob/a09d0448013911849be94f292c88db0ea4d8e7c8/src/lib/ws/middleware/probe-metadata.ts#L20

and that one is on the list, so I think that works as originally intended?

My point still is that it might be better to not drop probes forever, only apply some long-ish timeout.

alexey-yarmosh:
`failed to collect probe metadata` is replaced with `failed to collect probe metadata x.x.x.x` [here](https://github.com/jsdelivr/globalping/blob/a09d0448013911849be94f292c88db0ea4d8e7c8/src/lib/ws/middleware/probe-metadata.ts#L22-L24).

Size of delay surely can be set to any value, we just have 7 days by default.

MartinKolarik:
I see. Let's fix the reconnecting and reduce the timeout, then.

alexey-yarmosh:
Do you think it should be reduced for all fatalConnectErrors or just for unresolvable geo?

MartinKolarik:
I'd say all. 1 hour should be enough to prevent the load caused by constant reconnecting and, at the same time, reasonably short to fix any temporary failures.

alexey-yarmosh:
> I'd say all. 1 hour should be enough to prevent the load caused by constant reconnecting and, at the same time, reasonably short to fix any temporary failures.

Adding that logic revealed an #390 issue. So it will be implemented along with the fix. But current PR is safe to merge and deploy right now, so lets do that.

PR #403, opened 2023-07-25
feat: add docs links to error responses

MartinKolarik:
- Deletes the old docs.
 - Adds links to the new docs into all error responses (logic reused from the jsDelivr API):

```
{
  "error": {
    "type": "validation_error",
    "message": "Parameter validation failed.",
    "params": {
      "limit": "\"limit\" must be less than or equal to 500"
    }
  },
  "links": {
    "documentation": "https://www.jsdelivr.com/docs/api.globalping.io#post-/v1/measurements"
  }
}
```


docs/README.md
jimaek, line 1: We link to this file in many places. Maybe we shouldn't delete it, just update it to point to the new docs

MartinKolarik: I'll update the references I find, but we can do this as well, sure.

src/probe/route/get-probes.ts
```diff
@@ -37,5 +37,5 @@ const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & R
 };
 
 export const registerGetProbesRoute = (router: Router): void => {
-	router.get('/probes', handle);
+	router.get('/probes', '/probes', handle);
```
alexey-yarmosh, line 40: Why there should be name + route instead of just route?

MartinKolarik: The paths start to be messy when parameters get involved, so the router uses the name to refer to the routes internally: 

https://github.com/jsdelivr/globalping/pull/403/files#diff-3cb4878d5e7a24b1637229f8b009ba4736558babd22cf9cbbeeb6b373f69f260R27

If the route doesn't have a name, it can't be referenced, and `_matchedRouteName`, which we use here https://github.com/jsdelivr/globalping/pull/403/files#diff-ef7c9ff1935f2d36ee4af016c62171bc466fa6b570629c0f911410e4f319d18fR5 stays `undefined`, so you can't detect the current route and build a link based on it.

docs/measurement/schema/response.md
```diff
@@ -1,934 +0,0 @@
...
-
-#### probesCount
-
-**key**: `probesCount`
-
-**type**: `number`
-
-**mandatory**: `true`
-
-The number of probes that performed the measurement. `probesCount` is always smaller or equal to `limit`.
```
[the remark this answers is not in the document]
MartinKolarik: `probesCount` added. I intentionally didn't include descriptions for things that should be obvious because I think writing "id - the measurement id" just adds extra text to read with no value.

alexey-yarmosh (approved):

Issue #400, opened 2023-07-24
unresolvable geoip for one of the probes


alexey-yarmosh:
```
[2023-07-24 14:13:38] [ERROR] [56] [probe-metadata] unresolvable geoip: x.x.x.x
{ expose: true }
Error: unresolvable geoip: x.x.x.x
    at GeoipClient.lookup (file:///app/dist/lib/geoip/client.js:48:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async buildProbe (file:///app/dist/probe/builder.js:39:18)
    at async file:///app/dist/lib/ws/middleware/probe-metadata.js:11:32
[2023-07-24 14:13:38] [INFO] [56] [ws:error] disconnecting client for (unresolvable geoip: x.x.x.x) [x.x.x.x]
[2023-07-24 14:13:38] [DEBUG] [56] [ws:error] unresolvable geoip: x.x.x.x
{ data: { ipAddress: 'x.x.x.x' } }
Error: unresolvable geoip: x.x.x.x
    at file:///app/dist/lib/ws/middleware/probe-metadata.js:20:15
```

PR #407, opened 2023-07-26
fix: fix DC and FM states mapping

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/400


src/lib/location/states.ts
```diff
@@ -9,8 +9,8 @@ export const states: Record<string, string> = {
 	'Connecticut': 'CT',
 	'Delaware': 'DE',
 	'Washington, D.C.': 'DC',
-	'District Of Columbia': 'DC',
-	'Federated States Of Micronesia': 'FM',
+	'District of Columbia': 'DC',
+	'Federated States of Micronesia': 'FM',
```
MartinKolarik, line 13: I wonder why this one is even here, it's a country with its own ISO code, not a US state: https://ipinfo.io/119.252.112.0

alexey-yarmosh: Removed

MartinKolarik (approved):

Issue #354, opened 2023-05-06
Incorrect private IP detection


jimaek:
192.175.48.1 is a public IP but the API blocks it "[target] Private hostnames are not allowed."

PR #410, opened 2023-07-28
Refactor: replace `private-ip` with Node.js native API

SukkaW:
The PR replaces the `private-ip` npm package with Node.js built-in native API.

The `net.BlockList` API is first introduced in Node.js 14.18.0 and 15.0.0, 
and its API doc can be found here: https://nodejs.org/dist/latest-v18.x/docs/api/net.html#class-netblocklist.

The `net.BlockList` API is originally designed for the purpose of disabling inbound or outbound access to specific IP addresses, IP ranges, or IP subnets. However, it is also a high-performance IP List and can check whether an IP address is included in the list.

The PR creates a `net.BlockList` instance and adds every known private IP range to the list.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,6 @@
 				"maxmind": "^4.3.10",
 				"newrelic": "^10.4.1",
 				"physical-cpu-count": "^2.0.0",
-				"private-ip": "^3.0.0",
 				"rate-limiter-flexible": "^2.4.1",
 				"redis": "^4.6.5",
 				"request-ip": "^3.3.0",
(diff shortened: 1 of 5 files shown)


SukkaW:
I have noticed that the unit tests (in GitHub Action) have reported the following error:

```
[ERROR] [2991] [probe-metadata] couldn't detect probe location for ip 79.205.97.254
```

It appears that these errors may not be directly attributed to the changes made in my PR. Will it have any impact on mering my PR?

MartinKolarik:
Hi, thanks for the PR. Although I suggested a different change in #354, I'm happy to accept this one as it's likely to perform better.

The failing tests were caused by one of the IPv6 ranges. I'm not sure which one and why exactly, but since we don't support IPv6 at all yet, I deleted those rules for now.

Fixes #354.

MartinKolarik (approved):

Issue #379, opened 2023-06-15
Suburbs to city names


jimaek:
Related to https://github.com/jsdelivr/globalping/issues/264#issuecomment-1592913583

>A lot of mismatches happen because of preferring the suburb city instead of the main city (e.g. Dadri and Delhi) That is a point to think, do we want probes to be as accurate as possible (so user will be able to filter by Dadri, but filter by Delhi will not return that probe) or we want to accumulate all suburb cities in main city. Third option is to combine both and show that probe both in Dadri and Delhi searches. Other examples: Doddaballapura and Bangalore, Piscataway and New York.

>Top thing to do is to find the main city for district/suburb. We can try geonames.org API or similar tools for that. For example [here](https://www.geonames.org/8133088/spanga.html) under the Hierarchy tab we can see that Spanga refers to Stockholm.

jimaek:
Note: Ideally we should make probes quarriable by both suburb and main city. So a probe could be in London and Red Bridge at the same time

MartinKolarik:
Most recent details https://github.com/jsdelivr/globalping/issues/264#issuecomment-1629455226

PR #412, opened 2023-08-02
feat: city approximation

alexey-yarmosh:
Part of: https://github.com/jsdelivr/globalping/issues/264
Fixes: https://github.com/jsdelivr/globalping/issues/379

Here we are getting approximated city by location (lat, long) from providers, so city info from providers is not used.

Accuracy results can be found [here](https://docs.google.com/spreadsheets/d/1a-r_YBKsPxIeXpaB6rKZusrrqBDQ83Up6fN5e3uwIYY/edit#gid=33495994).

At first glance accuracy didn't change, but we fixed almost all suburb issues. Some other issues showed up, and here is what we can do:
- A lot of times location is changed to the city of roughly the same size nearby. "Piscataway" -> "Staten Island", "Gunzenhausen" -> "Ansbach", "Ashburn" -> "Germantown", "Manila" -> "Quezon City":
    - some of such cases may be fixed by increase of search radius (e.g. to 50km) to find the main city of a place;
    - for other we can add logic to prefer approximated city only if the population of original city is smaller than X and if the population of approximated city is Y times bigger than original's one.
- "Falkenstein" is moved to "Zwickau" constantly, not sure that is ok:
  - there can be an exception for it and other popular but small locations not to apply approximation, don't think that could be automated;
  - maybe its bound to the data we have, but seems that if at least one providers says its "Falkenstein" its really "Falkenstein" :)
- Other wrong results are where majority of providers say it is place A and DC says it place B. Not sure if DC may be inaccurate, but if 5 providers say its duck maybe its duck? :)


jimaek:
Ideally we need a solution that keeps probes with city set to Ashburn and Falkenstein. They are quite popular and it's realistic that people will search for them specifically.
Any suggestions?

alexey-yarmosh:
Germantown is only 2x bigger than Ashburn so that will work:
> for other we can add logic to prefer approximated city only if the population of original city is smaller than X and if the population of approximated city is Y times bigger than original's one.

Zwickau is 10x bigger than Falkenstein, so I think there should be an exception to keep "Falkenstein". It is really specific location so I can't imagine any common rule for it.

jimaek:
>Zwickau is 10x bigger than Falkenstein, so I think there should be an exception to keep "Falkenstein". It is really specific location so I can't imagine any common rule for it.

But we know for a fact that Falkenstein is the correct location. Having a logic the breaks correct locations is not ideal. 
What city do the DBs see when it comes to a probe in Falkenstein?

jimaek:
Looks like:
ipinfo Frankfurt
ip2loc Falkenstein
maxmind Falkenstein

So where does Zwickau come from? They're not even close to each other

alexey-yarmosh:
Zwickau is ~30km away. https://goo.gl/maps/1hCb6TzQAfErSxim9

alexey-yarmosh:
By exception for Falkenstein I mean smth like:
```js
if (providerCity == 'Falkenstein') {
  return providerCity
} else {
  return getApproximatedCity();
}
```
So only for Falkenstein we are not applying approximation and every provider with Falkenstein response will participate in voting using Falkenstein value.

MartinKolarik:
> But we know for a fact that Falkenstein is the correct location. Having a logic the breaks correct locations is not ideal. 
What city do the DBs see when it comes to a probe in Falkenstein?

Well that's exactly what this task is about. Falkenstein seems to be a super small city, in some countries it wouldn't even be considered a city, just a village. If we do any kind of approximation to large nearby cities, then it loses this comparison.

MartinKolarik:
Maybe don't approximate if the city is on this (or similar) list? https://www.datacentermap.com/datacenters/

alexey-yarmosh:
> Maybe don't approximate if the city is on this (or similar) list? https://www.datacentermap.com/datacenters/

Yeah, that makes sense. That will also disable a few more wrong approximations, and make result correct:
Manila -> Quezon City
Manassas -> Centreville

At the same time that disables approximation of Noida -> Delhi, which is wrong according to our table (expecting Delhi).

Also, since we are starting to use city names from providers, we need https://github.com/jsdelivr/globalping/issues/163 to fix naming inconsistencies before checking them against the DC cities list.

jimaek:
> Maybe don't approximate if the city is on this (or similar) list? https://www.datacentermap.com/datacenters/

Wouldn't that break lots of locations too? 

@alexey-yarmosh have you looked into detection of type of probes? Data center or home user. We could then limit the approximation to home users only

alexey-yarmosh:
At first glance ipinfo and fastly have useful info. They both detected 5 out of 6 home probes (marked them as "non-hosting"). While DC probes are marked as "hosting".

jimaek:
So what does everyone think? Would this be the most optimal way to get the most accurate locations?
Keep the existing PR logic and add exception for DC probes?

alexey-yarmosh:
Providers may return suburb cities + hosting: true, e.g.:
- ipinfo returns "Dadri" + hosting: true
- fastly retruns "Spanga" + hosting: true
- fastly retruns "Clifton" + hosting: true
- fastly retruns "Milpitas" + hosting: true

If we rely on `hosting: true` field all that values will participate in voting, which is wrong.
At the same time DC list doesn't have all that suburbs and won't prevent approximation, but will prevent in popular DC cities, which is exactly what we need.

Both solutions require naming normalisation anyway, so adding a list of popular DC cities is relatively easy.

alexey-yarmosh:
I've added a doc to describe the geo ip algorithm:
https://github.com/jsdelivr/globalping/blob/e391600d2203742e9481fb8eccea89e2fb306d3f/docs/geoip.md

jimaek:
I dont think we need to anonymize the providers. We can right down the exact names and priorities to make it easier to understand the logic and fix if needed.

alexey-yarmosh:
Potential changes if we add skipping approximation for cities from https://www.datacentermap.com/datacenters/ :

Result changed to the correct one:
Ashburn -> Germantown is now Ashburn (1 time)
Falkenstein -> Zwickau is now Falkenstein (2 times)
Manassas -> Centreville is now Manassas (1 time)
Manila -> Quezon City in now Manila (1 time)
total: 5 results fixed.

Result changed to the wrong one:
Noida -> Delhi is now Noida (1 time)
total: 1 result broke.

City changed for a single probe, but result is still wrong for another reason:
Piscataway -> Staten Island is now Piscataway
Santa Clara -> San Jose is now Santa Clara
Roubaix -> Lille is now Roubaix

So in total it is about +4% accuracy.

alexey-yarmosh:
Updated accuracy table can be found here:
https://docs.google.com/spreadsheets/d/1a-r_YBKsPxIeXpaB6rKZusrrqBDQ83Up6fN5e3uwIYY/edit#gid=832522498

tsconfig.json
```diff
@@ -2,6 +2,12 @@
 	"extends": "@tsconfig/node18-strictest-esm/tsconfig.json",
 	"compilerOptions": {
 		"outDir": "./dist",
+		"lib": [
+			"esnext"
+		],
+		"module": "esnext",
+		"target": "esnext",
```
alexey-yarmosh, line 9: `"module": "esnext"` is required to import json files, which I find very handy. Updating also `lib` and `module` to stay consistent.

src/lib/geoip/city-approximation.ts
```diff
@@ -0,0 +1,139 @@
...
+	return approximatedCity;
+};
+
+export const populateCitiesList = async () => {
+	redis = getRedisClient();
+	const cities = await readCitiesCsvFile() as CsvCityRow[];
+
+	geonamesCities = new Map(cities.map(city => ([ city.geonameId, { ...city, population: parseInt(city.population, 10) }])));
+
+	await Promise.all(cities.map(async (city: City) => {
```
MartinKolarik, line 78: It's better to use `Bluebird.map()` (http://bluebirdjs.com/docs/api/promise.map.html) with limited `concurrency` for looping over large lists, otherwise, 25k promises get spawn at the same time here, which may cause various random issues with the whole process.

Issue #418, opened 2023-08-10
Setup prod performance simulation


jimaek:
We need a way to test performance related changes on a system that has at least 800 probes connected and can scale to whatever number we need.

This way we can test stuff like Redis load, fetchSockets timeouts, Dragonflydb, future scaling to thousands of probes...

PR #423, opened 2023-09-01
feat: staging env set up

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/418

Here is the md doc + all changes required to run a prod performance simulation (staging env).


redis.conf
MartinKolarik, line 1: Let's put this into `config/redis.conf`

MartinKolarik (approved):

Issue #425, opened 2023-09-07
Adopting probes


MartinKolarik:
> Ability to adopt a probe. A user should be able to enter the IPs of the probes he controls to adopt them. I imagine this working like this: 1. Enter IP address of probe 2. Message asks the user to login where the probe is running and follow the logs output 3. The API sends a special command to that specific probe that prints a message to the user in the logs. It would contain a unique one-time code that would verify ownership 4. The user enters the code in the dashboard 5. The probe is now his

> We agreed that IPs will be used to identify probes. Once a probe is assigned to an account, it'll belong to that account even after reconnecting, as long as it has the same IP.
The IP ownership will expire after 30 days of being offline.
No extra limits for sponsors. Everything above the base limits is tracked as credits.

 - Adopted probes can be removed from the account by the user at any point. This also resets all custom metadata.

alissonlauffer:
I have a probe running on my home network and its IP changes every time I reboot the modem. How would it handle these cases?

jimaek:
You're right. We should use 2 parameters at the same time.
IP and a UUID generated on boot. And if at least one of them remains the same then the probe remains adopted. If both change then connection gets lost

alissonlauffer:
Couldn't the UUID be persistent? So as long as I don't delete the container data, it remains mine. Because with the UUID generated on boot approach, if I lose power and both devices reboot (thus resetting both my IP and my UUID), I'd lose the adoption (even though the container is the same).

jimaek:
Our probes are all read-only at the moment, so it wouldn't work as is.

The above solution should cover the majority of cases, and in the future we will definitely add a more advanced option for the case you described. Possibly reading the UUID from a file if the user is technical enough to mount it to the container, or using a USB stick in case of hardware probes.

PR #435, opened 2023-10-06
feat: adoption probes

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/425
This functionality requires a connection to MariaDB with the 'adopted_probes' table configured.


docker-compose.yml
```diff
@@ -8,3 +8,14 @@ services:
     volumes:
       - ./config/redis.conf:/usr/local/etc/redis/redis.conf
     command: /usr/local/etc/redis/redis.conf
+
+  mariadb:
+    image: mariadb:latest
```
MartinKolarik, line 13: We should match the prod version here.

src/adoption-code/schema.ts
```diff
@@ -0,0 +1,6 @@
+import Joi from 'joi';
+
+export const schema = Joi.object({
+	ip: Joi.string().ip().required(),
```
MartinKolarik, line 4: Should use `globalIpOptions`

config/init.sql
```diff
@@ -0,0 +1,19 @@
...
+  userId VARCHAR(255) NOT NULL,
+  ip VARCHAR(255) NOT NULL,
+  uuid VARCHAR(255),
+  lastSyncDate DATE
+);
+INSERT IGNORE INTO adopted_probes (id, userId, ip) VALUES ('1', '6191378', '79.205.97.254');
+
+CREATE USER IF NOT EXISTS 'directus';
+GRANT ALL PRIVILEGES ON `directus`.* TO 'directus';
+FLUSH PRIVILEGES;
```
MartinKolarik, line 19: This is dev only and in production, we rely on directus migrations, right?

src/lib/adopted-probes.ts
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	private async syncDashboardData () {
+		const allSockets = await this.fetchWsSockets();
+		this.connectedIpToUuid = new Map(allSockets.map(socket => [ socket.data.probe.ipAddress, socket.data.probe.uuid ]));
+		this.connectedUuidToIp = new Map(allSockets.map(socket => [ socket.data.probe.uuid, socket.data.probe.ipAddress ]));
+
+		const adoptedProbes = await this.sql(TABLE_NAME).select<AdoptedProbe[]>('ip', 'uuid', 'lastSyncDate');
+		await Promise.all(adoptedProbes.map(({ ip, uuid }) => this.syncProbeIds(ip, uuid)));
+		await Promise.all(adoptedProbes.map(({ ip, lastSyncDate }) => this.updateSyncDate(ip, lastSyncDate)));
```
MartinKolarik, line 40: Use `Bluebird.map()` with limited concurrency.

test/tests/unit/adopted-probes.test.ts
```diff
@@ -0,0 +1,155 @@
...
+		fetchSocketsStub.resolves([{ data: { probe: { ipAddress: '1.1.1.1', uuid: '1-1-1-1-1' } } }]);
+
+		adoptedProbes.scheduleSync();
+		await sandbox.clock.tickAsync(5000);
+
+		expect(whereStub.callCount).to.equal(0);
+		expect(updateStub.callCount).to.equal(0);
+		expect(deleteStub.callCount).to.equal(0);
+	});
+});
```
MartinKolarik, line 155: No point in rewriting it now but note that here I'd greatly prefer integration tests that prepare data in a real DB, run the sync, and check the results. Would be much cleaner than all the mocking and provide better coverage as well.

config/init.sql
alexey-yarmosh: yes, directus uses `directus` user so it will be in place in production. maybe we should also have a specific user for gp api?

MartinKolarik: User can be the same.

src/lib/adopted-probes.ts
```diff
@@ -0,0 +1,112 @@
...
+		private readonly sql: Knex,
+		private readonly fetchWsSockets: typeof fetchSockets,
+	) {}
+
+	scheduleSync () {
+		setTimeout(() => {
+			this.syncDashboardData()
+				.finally(() => this.scheduleSync())
+				.catch(error => logger.error(error));
+		}, 5000);
```
MartinKolarik, line 30: If we keep this running in multiple processes, let's bump the sync interval to 1 minute.

MartinKolarik:
Even though we don't directly test it, the process is trying to connect to the DB during tests. This previously didn't fail only because the timeout was 1 minute and the tests run faster then that:

image

Now, it fails after I reduced the connection timeout. We need to either prevent this or configure the DB on CI (https://github.com/jsdelivr/data.jsdelivr.com/blob/master/.github/workflows/nodejs.yml#L27-L32).

Issue #427, opened 2023-09-07
Setting data for adopted probes


MartinKolarik:
- Rename probes but limited to the user's point of view. Not public. (dashboard only)
 - User can change the probe city (and this updates the coordinates in the background). Must be a valid city in the detected country.
 - Tag system. Allow the user to set a limited amount of tags on adopted probes. Then he would be able to use the API to query them specifically. Something like tag:u-jimaek-value . Example use case: A company wants to use Globalping to run tests from their own specific locations and avoid all public probes. They would add them to the network and adopt them. They would then be shared by all of course. But they can also tag them to allow them to run their own tests only from their probes. e.g. a CDN running a probe in every POP to test their latency to the internet

alexey-yarmosh:
A few questions:
- are all user tags publicly visible to everyone?
- algorithm of creating `u-jimaek-value` tag. As I understand user only input `value` part, and `u-jimaek-` is autogenerated. To get the `jimaek` value:
  - dashboard BE does a gh api request by gh id during the creation of a new adoption, and stores login in the adopted probe item;
  - since gh login may change we also need a cron that will sync gh logins by gh ids.
- validating the city. I see the algorithm as:
  - user edits the adopted probe and fulfils 2 inputs: country and city;
  - dashboard BE does a geonames request with the country and city values and saves the city, country and coordinates of the best result in the adopted probe item;
  - GP API only uses city from the adoption if country from adoption === country from geoip search.

geonames has is 10000 credits daily limit per application, the hourly limit is 1000 credits. 1 search requires 1 credit.

@MartinKolarik @jimaek does that look ok?

jimaek:
> A few questions:
> 
> * are all user tags publicly visible to everyone?

Yes
> * algorithm of creating `u-jimaek-value` tag. As I understand user only inputs the `value` part, and `u-jimaek-` is autogenerated. To get the `jimaek` value:
>   
>   * dashboard BE does a gh api request by gh id during the creation of a new adoption, and stores login in the adopted probe item;

We get the username during registration, no need to make new requests during adoption. Just take the value from our DB.
>   * since gh login may change we also need a cron that will sync gh logins by gh ids.

I dont think we need a cron, we can auto-update usernames on every login into the dashboard
> * validating the city. I see the algorithm as:
>   
>   * user edits the adopted probe and fulfils 2 inputs: country and city;

We already have these 2 fields filled with system data. Why do we need to make the country editable at all?

[...]

Issue #268, opened 2022-12-28
For users: Report wrong IP geo location here


jimaek:
If you are hosting a probe and know that the resolved location is wrong:

1) If the city is wrong but the country is correct, you can set the correct city yourself after adopting the probe on our [dashboard](https://dash.globalping.io/).

2) If the country is wrong, please report it here. We will use this data to improve our algorithm. Please do it only if you're fine with making your IP public.

   Please comment with:

   ```
   Real location:
   Resolved location:
   IP address:
   ```

yoursunny:
Real location: Dallas TX, USA
Resolved location: "(New Zion, US, NA) (lat: 33.7957 long: -80.0148)"
IP address: 108.174.203.*

This is LigaHosting VPS, described to be in Dallas, confirmed via ping latency.

gshapiro:
Real location: Milpitas, US, NA
Resolved location: Santa Clara, US, NA
IP address: 98.51.131.164

It was close (within 8 miles), not sure what fidelity you are looking for.  The probe reported:

[2023-01-13 22:32:50] [INFO] [695] [api:connect] connected from (Santa Clara, US, NA) (lat: 37.3541 long: -121.9552)

Ne00n:
Real Location: FFM
Resolved Location: BE
Probe: Berlin, DE, EU, 210993

skapytek:
Real Location: Wolfsegg am Hausruck, Upper Austria, Austria
Resolved Location: Laakirchen
Probe: Laakirchen, AT, EU, 8412 -- T-Mobile Austria GmbH
If its not a problem i wont write my IP publicly cause ive installed the probe at my privat home.

skapytek:
Real Location: Wolfsegg am Hausruck, Upper Austria, Austria
Resolved Location: Laakirchen
Probe: Laakirchen, AT, EU, 8412 -- T-Mobile Austria GmbH
IP:  84.113.52.3

jimaek:
We released a new version of our geo IP resolution logic. I invite everyone to check their probes and provide us with feedback.

gshapiro:
> We released a new version of our geo IP resolution logic. I invite everyone to check their probes and provide us with feedback.

@jimaek What sort of accuracy are you interested in having reported?  My probe geolocation is within 3 miles of actual (though different city).

jimaek:
3 miles is considered good for us :) Once we release our dashboard you will be able to correct it manually too

gshapiro:
[...]

PR #442, opened 2023-11-10
feat: setting data for adopted probes

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/427


MartinKolarik (commented):
Not finished with the review yet, but submitting what I have so far.

src/lib/geoip/utils.ts
```diff
@@ -9,6 +9,8 @@ export const normalizeCityNamePublic = (name: string): string => {
 
 export const normalizeCityName = (name: string): string => normalizeCityNamePublic(name).toLowerCase();
 
+export const normalizePublicName = (name: string): string => name.toLowerCase();
```
MartinKolarik, line 12: Probably related to the question on the dashboard side - do we need this?

test/tests/unit/probe/router.test.ts
```diff
@@ -819,5 +819,26 @@ describe('probe router', () => {
 
 			expect(probes.length).to.equal(0);
 		});
+
+		it('should return match for user tag', async () => {
```
MartinKolarik, line 823: Let's add a test for the part that user tags are not used in magic mode.

src/lib/adopted-probes.ts
```diff
@@ -1,49 +1,122 @@
...
+			shouldUpdateIfCustomCity: true,
+		},
+		network: {
+			connectedField: 'location.network',
+			shouldUpdateIfCustomCity: true,
+		},
+		country: {
+			connectedField: 'location.country',
+			shouldUpdateIfCustomCity: false,
+		},
```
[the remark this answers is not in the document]
alexey-yarmosh: If countries of our geo resolution and adoption are different - adoption data is not applied. So result will be C2 with the city from geo resolution logic.
At the same time adopted probe city/country data is not changed (just not used). So if country goes back to C1 city will be A2 again.

src/lib/geoip/utils.ts
alexey-yarmosh: answered here: https://github.com/jsdelivr/dashboard/pull/9#discussion_r1395520408

test/tests/unit/probe/router.test.ts
alexey-yarmosh: Since router.test.ts works with already build sockets I've added a new test to the create-measurement.test.ts + building is covered in a new fetch-sockets.test.ts

src/lib/adopted-probes.ts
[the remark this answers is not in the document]
MartinKolarik: I think we should update the country in the DB to C2 (but keep the city as set by the user) so that the user can readjust the city in C2 if needed.

src/lib/geoip/utils.ts
MartinKolarik: Ok, so on the dashboard side, we need to apply the normalization so that the values are presented to the user in the same way as from the API. I still don't understand the purpose of this function here. It's called `normalizePublicName()` but lowercasing is otherwise included only in the "internal" normalization, not public.

src/lib/adopted-probes.ts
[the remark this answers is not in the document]
alexey-yarmosh: Is that ok to display city not from the displayed country on the dashboard?

Plus, what is the use case for that? User changes the city to the correct A2 in C1. But when geo logic says it is another country C2, user founds another correct city?

src/lib/geoip/utils.ts
alexey-yarmosh: Function takes "public name" as argument, and normalizes it => `normalizePublicName`. Similar to `normalizeRegionName` or `normalizeNetworkName`.

MartinKolarik: Oh, I see, it isn't "normalize *to* public name" (what other functions there do) but "normalize *from* public name to internal one". You could probably just use `normalizeCityName()`, it'll repeat part of the process but that shouldn't cause any issues.

src/lib/adopted-probes.ts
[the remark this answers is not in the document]
MartinKolarik: > Is that ok to display city not from the displayed country on the dashboard?

I think that's exactly what we need. If the dashboard says C1, but the probe actually can't be targeted via C1 because API sees C2, it'll look broken and confusing. The dashboard should always show the current state.

> user founds another correct city?

Could be the IP got moved... But I agree that's very rare. More likely is that C2 is wrong, and in that case the previous point is important - it should be visible in the dashboard otherwise it'll be confusing.

src/lib/geoip/utils.ts
alexey-yarmosh: This function will be called on every fetchSockets for every adopted probe so it will be great to avoid regex operations there, I can just call `toLowerCase()` in that single place where the function was used.

MartinKolarik: Let's rename to `normalizeFromPublicName` then and move on, that should be more clear.

src/lib/adopted-probes.ts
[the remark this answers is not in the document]
alexey-yarmosh: > If the dashboard says C1, but the probe actually can't be targeted via C1 because API sees C2, it'll look broken and confusing

Same can be applied to cities. 

We can add something like `editIsActive` flag to indicate status of custom city, but that may confuse even more.

I personally think not applying change if country is different without updates on dashboard is the simples logic to understand.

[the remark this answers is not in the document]
MartinKolarik: >Same can be applied to cities.

Yes, it would be best to show a warning in that case as well. Otherwise, it won't be clear to anyone what happened when it doesn't work.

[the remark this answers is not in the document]
alexey-yarmosh: So we are only sending notification (similar to gh username change notification) and ignoring city and country from the adoption? Or additionally changing the country on the dashboard?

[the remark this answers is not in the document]
MartinKolarik: Let's change country in dashboard, notify and link to https://github.com/jsdelivr/globalping/issues/268 in the notification (If this change is not right, you can report it at ...).

alexey-yarmosh:
Updated to handle the adopted country change, @MartinKolarik please review.

MartinKolarik (approved):

PR #448, opened 2023-11-27
feat: add truncated http field

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/357


public/v1/components/schemas.yaml
```diff
@@ -1056,6 +1056,9 @@ components:
             rawBody:
               type: string
               description: The raw HTTP response body. Only the first 10 kb are returned.
+            truncated:
```
MartinKolarik, line 1059: We can also add it to the `required` list above since it's always present (which might trigger some extra validations and force you to add it to tests :smile:).

Issue #430, opened 2023-09-08
Allow reusing measurement IDs for selecting same group of probes


jimaek:
Fixes #363 

This feature will allow non-stop tests for ping and mtr (supported by client logic) and also allow anyone to compare 2 endpoints using the exact same probes without knowing their IDs.

The measurement ID to probe mapping must be instant to allow the user to start a new measurement with the same probes while the first one is still running,

MartinKolarik:
In terms of the API, I'm thinking of overloading the `locations` array so that if you pass in a string, it's interpreted as a measurement ID, and the locations are reused from that ID.

alexey-yarmosh:
measurement1 with id1 had 20 probes, user starts a new measurement with `locations: id1` but API found only 19 of prev probes (1 is disconnected right now). Should api proceed and return `OK {probesCount: 19}` or try replace missing or send 422?

alexey-yarmosh:
`offline` status makes sense. Then, since it is treated as a probe with error, probesCount should be 20 I think.

MartinKolarik:
Yes.

PR #453, opened 2023-12-05
feat: allow reusing measurement IDs for selecting same group of probes

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/430


src/lib/geoip/providers/ip2location.ts
```diff
@@ -59,11 +60,11 @@ export const ip2LocationLookup = async (addr: string): Promise<Ip2LocationBundle
 		longitude: result.longitude ?? 0,
 		network: result.as ?? '',
 		normalizedNetwork: normalizeNetworkName(result.as ?? ''),
+		isHosting: result.usage_type ? HOSTING_USAGE_TYPES.includes(result.usage_type) : null,
 	};
 
 	return {
 		location,
-		isHosting: result.usage_type ? HOSTING_USAGE_TYPES.includes(result.usage_type) : undefined,
```
MartinKolarik, line 66: No objections here, just curious about the reason for this change.

src/measurement/store.ts
```diff
@@ -155,13 +181,16 @@ export class MeasurementStore {
 
 	removeDefaults (measurement: MeasurementRecord, request: MeasurementRequest): Partial<MeasurementRecord> {
 		const defaults = getDefaults(request);
+
 		// Removes `"limit": 1` from locations. E.g. [{"country": "US", "limit": 1}] => [{"country": "US"}]
-		measurement.locations = measurement.locations.map(location => location.limit === 1 ? _.omit(location, 'limit') : location);
+		if (_.isArray(measurement.locations)) {
+			measurement.locations = measurement.locations.map(location => location.limit === 1 ? _.omit(location, 'limit') : location);
+		}
```
MartinKolarik, line 188: Actually, here, we should probably inject the original locations 🤔 Otherwise clients will need to implement special logic to handle this case and fetch the original measurement themselves. Maybe store it in the same object as the IPs?

src/probe/sockets-location-filter.ts
```diff
@@ -1,153 +0,0 @@
-import config from 'config';
```
MartinKolarik, line 1: Small note here - when renaming files, it's really good to commit the rename on its own and then make any changes separately so that git diff understands it's the same file and doesn't show it as one full delete and second full add (which is hard to review).

src/probe/types.ts
```diff
@@ -47,3 +46,31 @@ export type Probe = {
...
+	version: null;
+	nodeVersion: null;
+	uuid: null;
+	isHardware: false;
+	hardwareDevice: null;
+	ipAddress: string;
+	host: null;
+	index: [];
+	tags: {
+		type: 'offline';
```
MartinKolarik, line 64: What's the purpose of this?

test/tests/unit/measurement/store.test.ts
```diff
@@ -223,103 +331,57 @@ describe('measurement store', () => {
...
+							city: 'city',
+							asn: 'asn',
+							longitude: 'longitude',
+							latitude: 'latitude',
+							network: 'id',
+							tags: [],
+							resolvers: [],
+						},
+						result: {
+							status: 'failed',
```
MartinKolarik, line 376: Do you think it isn't worth creating its own status value in the end?

src/lib/geoip/providers/ip2location.ts
alexey-yarmosh: `ProbeLocation` type already has `isHosting` field, so no need to store it separately.

alexey-yarmosh: Ahah 😄. It is actually there in master: https://github.com/jsdelivr/globalping/blob/master/src/probe/types.ts#L13
I moved it from `ProbeLocation` to `LocationInfo`. Idea is to have one type for location that provider is returning, and one for location that is stored in probe. Ideally `isProxy` should be moved to `LocationInfo` too with default `null` value. So mutants like `Ip2LocationBundledResponse` can be removed.

src/measurement/store.ts
alexey-yarmosh: `redis.json.get` to the original measurement `locations` should work here.

test/tests/unit/measurement/store.test.ts
alexey-yarmosh: Initially I thought yes. But maybe for some UI with a list of probes constantly pinging it will be nice to mark either probe is offline or measurement failed. In that case make a condition based on `rawOutput` isn't nice, so introducing `status: offline` makes sense.

alexey-yarmosh:
I think we should additionally inform users that while they can create measurement without waiting for the prev to finish, measurement `id`s are not stored forever and shouldn't be hardcoded. Instead they should be rotated, ideally after every measurement finish.

src/probe/types.ts
alexey-yarmosh: Yes

MartinKolarik:
Also I think this is missing?

> Additionally, if locations.length === 1 and it's magic, and it doesn't match anything, as the last step, we should try to interpret it as a measurement ID as well. This will allow users to paste the ID into any of our tools that support the magic mode.

src/measurement/store.ts
MartinKolarik: Yes.

MartinKolarik (approved):

Issue #451, opened 2023-12-04
Fix DB setup in tests


MartinKolarik:
The tests must set up the expected DB state before running, not only in the CI as now, but also locally (from the node process). See https://github.com/jsdelivr/data.jsdelivr.com/blob/master/test/setup-db.js for a good example.

PR #458, opened 2023-12-13
fix: fix db setup in tests

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/451:
- init.sql moved to /migrations. It is used in docker-compose to prepare db in dev env, plus in tests setup to prepare test env.
- tests also need seeds data, so there is seeds/test/index.js


test/setup.ts
```diff
@@ -10,24 +12,39 @@ import {
...
-
 import chaiOas from './plugins/oas/index.js';
 import { getRedisClient, initRedis } from '../src/lib/redis/client.js';
-import { client } from '../src/lib/sql/client.js';
-import { USERS_TABLE } from '../src/lib/adopted-probes.js';
+import { client as sql } from '../src/lib/sql/client.js';
+
+if (process.env['NODE_ENV'] !== 'test') {
+	throw new Error(`NODE_ENV is not 'test' but '${process.env['NODE_ENV']}'.`);
+}
```
MartinKolarik, line 21: Let's keep the check as it was in the other repo. Env itself doesn't mean much.

```js
if (!dbConfig.connection.database.endsWith('-test') && dbConfig.connection.host !== 'localhost') {
	throw new Error(`Database name for test env needs to end with "-test" or the host must be "localhost". Got "${dbConfig.connection.database}"@"${dbConfig.connection.host}".`);
}
```

alexey-yarmosh: Do you want to divide `connection: 'mysql://directus:password@localhost:3306/directus',` on separate fields in config, or check for `connection !== 'mysql://directus:password@localhost:3306/directus'` in setup.ts?

MartinKolarik (approved):

Issue #462, opened 2023-12-22
Default uppercase `World` does not work on website


MarvinJWendt:
Hi, by default, the `Location` field on the website is set to `World`. This results in only a single selected probe. If I manually set it to `world`, it works as expected:

## Default (uppercase `World`)
![image](https://github.com/jsdelivr/globalping/assets/31022056/67cadba0-f637-4b83-bb7b-37a1790b2020)

## Manually changed to `world`
![image](https://github.com/jsdelivr/globalping/assets/31022056/11af6aac-10a1-4846-a968-5df98ed1c0b3)

jimaek:
Thanks for reporting, it will be deployed during the weekend

PR #463, opened 2023-12-22
fix: filter by world in any case

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/462

### DIFF
--- a/src/probe/probes-location-filter.ts
+++ b/src/probe/probes-location-filter.ts
@@ -59,7 +59,7 @@ export class ProbesLocationFilter {
 	}
 
 	public filterByLocation (probes: Probe[], location: Location): Probe[] {
-		if (location.magic === 'world') {
+		if (location.magic?.toLowerCase() === 'world') {
 			return _.shuffle(this.filterGloballyDistibuted(probes, probes.length));
 		}
 
--- a/test/tests/integration/measurement/create-measurement.test.ts
+++ b/test/tests/integration/measurement/create-measurement.test.ts
@@ -334,6 +334,22 @@ describe('Create measurement', () => {
 				});
 		});
 
+		it('should create measurement with "magic: World" location in any case', async () => {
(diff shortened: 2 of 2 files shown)


src/probe/probes-location-filter.ts
```diff
@@ -59,7 +59,7 @@ export class ProbesLocationFilter {
 	}
 
 	public filterByLocation (probes: Probe[], location: Location): Probe[] {
-		if (location.magic === 'world') {
+		if (location.magic?.toLowerCase() === 'world') {
```
MartinKolarik, line 62: This would still throw if `magic` is undefined. Needs to be `location.magic?.toLowerCase?.()` to have any effect.

alexey-yarmosh: `magic: undefined` is not possible from the app logic since we have `exactOptionalPropertyTypes: true`, and such case will break further anyway because we use constructions like `Object.keys(location).forEach()` and `Object.keys(location).includes('magic')`.

MartinKolarik: Well then why is there the first `?.` if it's not possible?

alexey-yarmosh: `{someOtherField: 'value'}` is possible `location.magic?.toLowerCase()` check is for that case. 
Explicit `{someOtherField: 'value', magic: undefined}` is not possible.

alexey-yarmosh: Seems in any of the undefined cases expression will not get to `toLowerCase()`, so second `?` is not required.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#short-circuiting

MartinKolarik: Oh indeed, I confused this with another case 👍

MartinKolarik (approved):

PR #481, opened 2024-01-23
fix: add git and optimize dockerfile

MartinKolarik:


Dockerfile
```diff
@@ -1,13 +1,26 @@
...
 
-COPY . /app
+ADD package.json package-lock.json /app/
 WORKDIR /app
 RUN npm ci --include=dev
+COPY . /app
 RUN npm run build
 
+FROM node:18-bullseye-slim
+RUN apt-get update -y && apt-get install tini util-linux curl -y
```
jimaek, line 13: It doesn't need git when running?

MartinKolarik: No

jimaek (approved):

Issue #473, opened 2024-01-18
Separate limits for authenticated users


MartinKolarik:
Part of https://github.com/jsdelivr/dashboard/issues/18

 - [x] Rename the current "rate limit" mentions in config/code to "anonymous rate limit", e.g., `measurement.rateLimit` -> `measurement.anonymousRateLimit`. It will keep working as it does but apply only to unauthenticated requests.
 - [x] Add a new `measurement.authenticatedRateLimit` option with the default value 250.
 - [x] Users can send a header in the format `Authorization: Bearer TOKEN`
   - [x] If the token doesn't exist, is expired, or fails the origin check, the response is a `401` error.
   - [x] If the token is valid, the anonymous rate limit is not applied; instead, the authenticatedRateLimit applies with the user account id being used as the key.
 
Requirements:

 - new tokens must work instantly, without any delay
 - deleted/revoked tokens should stop working reasonably fast
 - don't query the DB on every single request
 - the first time the token is used each day, its `date_last_used` value is updated in the DB

Suggestion:

 - query the DB once a minute for all tokens; for each token found, store it in process memory as valid for the next two minutes
 - if a request comes with a token that isn't in memory - query the DB for the specific token and store the result - valid/invalid - for the next two minutes

Note that the same token will also make it possible to use credits if the user reaches their hourly quota, but I'll open a separate task for that.

I've made some small changes related to this in https://github.com/jsdelivr/globalping/tree/gh-473 so please continue there.

PR #483, opened 2024-01-26
feat: separate limits for authenticated users

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/473


src/lib/http/middleware/authenticate.ts
```diff
@@ -0,0 +1,28 @@
...
+		const token = parts[1]!;
+		const origin = ctx.get('Origin');
+		const userId = await auth.validate(token, origin);
+
+		if (!userId) {
+			ctx.status = 401;
+			return;
+		}
+
+		ctx['auth'] = userId;
```
MartinKolarik, line 24: Not really sure if it has any impact but the recommended place for storing this is `ctx.state` so I'd suggest `ctx.state['userId']` here.

MartinKolarik (commented):
Very nice 👌

MartinKolarik (approved):

Issue #421, opened 2023-08-29
Possibility to get rid of commonjs modules


alexey-yarmosh:
Seems https://github.com/newrelic/node-newrelic/releases/tag/v11.0.0 may help us to get rid of requiring modules in commonjs style (https://github.com/jsdelivr/globalping/blob/master/src/cjs-dependencies.cjs). Need to check that and update if applicable.

PR #490, opened 2024-02-09
fix: newrelic logs and commonjs modules

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/477
fixes https://github.com/jsdelivr/globalping/issues/421


MartinKolarik:
If nothing major is blocking this, let's change the target to master and merge before #489.

MartinKolarik (approved):

PR #495, opened 2024-02-18
perf: use a separate redis channel for each probe

MartinKolarik:
Uses a new feature of the adapter I added in https://github.com/MartinKolarik/socket.io-redis-adapter/tree/dynamic-private-2, which allows creating separate channels for private rooms (the original adapter only supports this for public rooms so far; if this works well, I'll make a PR to upstream).


MartinKolarik:
Looks like this reduced the redis load by another 50 - 60%

PR #496, opened 2024-02-20
feat: hardware adopted fields

alexey-yarmosh:
Fixes https://github.com/jsdelivr/dashboard/issues/32


src/lib/adopted-probes.ts
```diff
@@ -179,6 +184,7 @@ export class AdoptedProbes {
 		const isCustomCity = adoptedProbe.isCustomCity;
 
 		if (!connectedProbe) {
+			await this.updateProbeData(adoptedProbe, { status: 'offline' });
```
MartinKolarik, line 187: This seems like it will run an update every time we call this function even if it's already set to `offline`? 

Also, other status changes are already propagated correctly (e.g. QA fail)?

alexey-yarmosh: Fixed regular "offline" updates. Other statuses are propagated correctly.

MartinKolarik (approved):

Issue #503, opened 2024-03-11
Allow DNS requests for TLDs and root zone


kdrenard:
Trying a DNS query for an SOA record for ".com" or "." (root zone) and getting "Provided target is not a valid domain name".

Thanks!

alexey-yarmosh:
@jimaek should that be allowed only for `dns` type and for any record type?

PR #510, opened 2024-03-29
feat: allow DNS requests for TLDs and root zone

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/503


MartinKolarik (approved):

test/tests/unit/measurement/schema/schema.test.ts
```diff
@@ -1129,24 +1144,23 @@ describe('command schema', async () => {
 		it('should pass (tld)', () => {
 			const input = {
 				type: 'dns',
-				target: '.com',
+				target: 'com',
```
alexey-yarmosh, line 1147: OP of an issue used ".com" as a value for tld. Shouldn't that be supported as well? @MartinKolarik

MartinKolarik: I know, but that was probably a typo. It's either `com` or `com.` in the spec and all the existing tools (try `dig`).

Issue #502, opened 2024-03-10
Add a periodic check for "1 probe per IP" rule


MartinKolarik:
To do after #501.

Even though we already have a check for allowing just one probe per IP, it can fail in some special cases, such as network failures in retrieving the remote probes or a simple race condition when two clients connect each to a different server at once.

Given the distributed nature of the system, we can't really make the check 100% reliable. I suggest we simply add an extra periodic check and if we detect two probes with the same IP there, all except the first one get disconnected.

PR #513, opened 2024-04-04
feat: periodic ip limit check

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/502

### DIFF
--- a/src/lib/ws/helper/probe-ip-limit.ts
+++ b/src/lib/ws/helper/probe-ip-limit.ts
@@ -27,21 +27,29 @@ export class ProbeIpLimit {
 	}
 
 	async syncIpLimit () {
-		const sockets = await this.fetchRawSockets();
-		// Sorting sockets by id, so all workers will treat the same socket as "first".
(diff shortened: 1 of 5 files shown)


src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -1,19 +1,62 @@
...
+			this.syncIpLimit()
+				.finally(() => this.scheduleSync())
+				.catch(error => logger.error(error));
+		}, 60_000 * 2 * Math.random() * numberOfProcesses).unref();
 	}
 
-	const probes = await fetchProbes({ allowStale: false });
-	const previousSocket = probes.find(p => p.ipAddress === ip && p.client !== socketId);
+	async syncIpLimit () {
+		const sockets = await this.fetchRawSockets();
```
MartinKolarik, line 30: How about we use `fetchProbes()` here (even `allowStale: true` should be fine I think), and only call `fetchRawSockets()` if we need the socket instance (to disconnect it)?

MartinKolarik (approved):

Issue #536, opened 2024-08-09
Check random e2e failures


MartinKolarik:
https://github.com/jsdelivr/globalping/actions/runs/10320878547/job/28572446539#step:6:1074
https://github.com/jsdelivr/globalping/actions/runs/10321158227/job/28573377490#step:6:436

Reproducible (in a few attempts) locally. Suspecting https://github.com/jsdelivr/globalping/blob/6934953f8667740712e208802148c4928205aac6/src/lib/ws/helper/reconnect-probes.ts#L13 based on the logs.

alexey-yarmosh:
Main reason indeed was probe reconnection logic, it is now disabled in e2e.
Also, 1 time I got probe stuck on `Checking for the latest version`, may be it was just a connection problem, but if repeated should be addressed too:
https://github.com/jsdelivr/globalping/actions/runs/10816945087/job/30009228800#step:6:335

Issue #328, opened 2023-04-07
Limits endpoint


jimaek:
We need an endpoint that the user and all tools could call to see the limits that apply to the user.

E.g. anon would see something like "max:100, used:50, available:50, type:IP"

Later we would expand it to understand API keys and show remaining credits. But the format would be different so need a way to combine them.

MartinKolarik:
I'm thinking this format, where:
 - `type` is `ip` or `user`,
 - `read` is returned only if it applies (doesn't right now, might to anon users after #526),
 - `credits` section is returned if the request is authenticated

```json
{
  "rateLimit": {
    "measurements": {
      "create": {
        "type": "ip",
        "limit": 500,
        "consumed": 150,
        "remaining": 350,
        "reset": 1240
      },
      "read": {
        "type": "ip",
        "limit": 25000,
        "consumed": 150,
        "remaining": 350,
        "reset": 1240
      }
    }
  },
  "credits": {
    "remaining": 1240
  }
}
```

PR #543, opened 2024-09-11
feat: /limits endpoint

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/328
Fixes https://github.com/jsdelivr/globalping/issues/536


MartinKolarik (approved):
I realized the `consumed` field doesn't actually make sense here because, in the headers, its meaning is related to the request (the number of rate limit points consumed by the request). Total consumed points can be calculated easily so I removed it.

Issue #526, opened 2024-06-22
Rate limits for GET


MartinKolarik:
We're getting way too many GET requests from some users, e.g., 60 / second for a single measurement. With several measurements running, it adds up quickly. The measurements themselves seem valid, so I think this is more likely an oversight on the users' end than an intentional DoS attempt.

I suggest we implement some (very high) limits on the currently unlimited GET requests. They shouldn't impact regular users at all but should catch extreme cases like this.

Some options I'm currently considering (may be combined):

- 3 requests per measurement per (IP or authed user) per second. This means people don't get restricted when running many measurements at once but can't send too many requests with just 1 measurement.
- 25 000 requests / hour / IP - only for anonymous requests. This would additionally prevent intentional DoS attempts. The number is based on 250 free tests and 100 requests per each, which is way more than needed. No hourly limit when using an API key.

MartinKolarik:
Let's start with the first option but with these values:
 - 5 requests / 2 seconds / measurement / user or IP
 - `blockDuration` 5 seconds
 - error 429 when blocked, with `Retry-After` header with the number of seconds remaining

PR #549, opened 2024-09-27
feat: rate limits for GET

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/526

### DIFF
--- a/src/measurement/route/get-measurement.ts
+++ b/src/measurement/route/get-measurement.ts
@@ -1,8 +1,6 @@
 import type { DefaultContext, DefaultState, ParameterizedContext } from 'koa';
 import type Router from '@koa/router';
 import { getMeasurementStore } from '../store.js';
-import { corsAuthHandler } from '../../lib/http/middleware/cors.js';
-import { authenticate } from '../../lib/http/middleware/authenticate.js';
 import { getMeasurementRateLimit } from '../../lib/rate-limiter/rate-limiter-get.js';
 import createHttpError from 'http-errors';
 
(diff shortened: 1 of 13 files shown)


MartinKolarik (commented):
After some more thought, I think this limit should always be applied per IP address. The effect will be more or less the same, but there is a good reason we had CORS configured to disallow sending the tokens on GET. Doing so triggers additional preflight requests, which add extra load, and also makes it much harder to add a CDN caching for finished measurements in the future.

src/measurement/route/get-measurement.ts
```diff
@@ -15,14 +17,13 @@ const handle = async (ctx: ParameterizedContext<DefaultState, DefaultContext & R
 	const result = await store.getMeasurementString(id);
 
 	if (!result) {
-		ctx.status = 404;
-		return;
+		throw createHttpError(404, `Couldn't find the requested item.`, { type: 'not_found' });
```
MartinKolarik, line 20: Why? There's no reason to throw these where they can be set directly, and the extra message doesn't have much value either.

alexey-yarmosh: This is because of message. It is out of sync with the doc: https://github.com/jsdelivr/globalping/blob/get-rate-limiter/public/v1/components/responses.yaml#L66

alexey-yarmosh: Maybe lets use `Couldn't find the requested measurement.`? To differentiate missing measurement and wrong path.

MartinKolarik: Ok sure

MartinKolarik (approved):

Issue #537, opened 2024-08-17
Support IP ranges whitelist


jimaek:
https://github.com/jsdelivr/globalping/blob/master/config/whitelist-ips.txt
e.g. 8.8.8.8/24

PR #568, opened 2024-10-31
feat: whitelist ranges

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/537


jimaek:
Maybe lets also add an example IP range to the config to make it clear that its supported, e.g. 8.8.4.0/24

MartinKolarik (approved):

Issue #572, opened 2024-11-18
Dashboard probe sync issues


mfld-pub:
![image](https://github.com/user-attachments/assets/6da5fddd-3b97-447d-aa73-b96aec759914)

Checking probe logs I can see them actively doing work. 

I can also 

```ping globalping ping 8.8.8.8 -F <my tag> -L 26```

 and I see every one of them returning. 

I am certain I do not have 26 probes on 4 continents offline right now. 

Something is glitching.

MartinKolarik:
They should be online now but I'll keep this open until we fix the underlying cause.

mfld-pub:
22 on 4 off now. I will check if those 4 are really offline.

PR #574, opened 2024-11-20
fix: adopted probes sync issue

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/572

What it does:
- in case of duplication error, extra adoptions are deleted, then update repeats;
- individual sync errors are logged and resolved, not to break syncing of others;
- uuid equality check is removed as not required.

### DIFF
--- a/src/lib/override/adopted-probes.ts
+++ b/src/lib/override/adopted-probes.ts
@@ -15,7 +15,7 @@ const logger = scopedLogger('adopted-probes');
 
 export const ADOPTED_PROBES_TABLE = 'gp_adopted_probes';
 export const NOTIFICATIONS_TABLE = 'directus_notifications';
-const ER_DUP_ENTRY_CODE = 1062;
+export const ER_DUP_ENTRY_CODE = 1062;
 
 export type AdoptedProbe = {
 	id: string;
--- a/src/lib/override/adopted-probes.ts
+++ b/src/lib/override/adopted-probes.ts
@@ -216,11 +216,9 @@ export class AdoptedProbes {
 	}
 
 	private async resolveIfError (pr: Promise<void>): Promise<void> {
-		const result = await Bluebird.resolve(pr).reflect();
-
-		if (result.isRejected()) {
-			logger.error(result.reason());
-		}
+		return pr.catch((e) => {
+			logger.error(e);
+		});
 	}
 
 	private async fetchAdoptedProbes () {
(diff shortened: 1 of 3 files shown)


src/lib/override/adopted-probes.ts
```diff
@@ -376,6 +403,23 @@ export class AdoptedProbes {
 		}
 	}
 
+	private async deleteDuplicates (currentAdoptedIp: string, connectedProbe: Probe) {
+		const duplicates = await this.sql(ADOPTED_PROBES_TABLE)
+			.where({ ip: currentAdoptedIp })
+			.orWhere({ ip: connectedProbe.ipAddress })
+			.orWhere({ uuid: connectedProbe.uuid })
+			// First item will be preserved, so we are prioritizing probes with a custom city and online probes.
+			.orderByRaw(`isCustomCity DESC, lastSyncDate DESC, onlineTimesToday DESC, FIELD(status, 'ready') DESC`)
```
MartinKolarik, line 412: I initially thought about considering tags here as well, but now I'm thinking: for this case to happen, the user must have adopted the same probe twice. It makes most sense to simply use his most recent data, doesn't it? If the user, e.g., didn't set a custom city the second time, it would be weird if the city suddenly reappeared. So maybe just `date_updated DESC, date_created DESC` ?

src/lib/override/adopted-probes.ts
```diff
@@ -338,12 +349,28 @@ export class AdoptedProbes {
...
-			ip: connectedProbe.ipAddress,
-			altIps: JSON.stringify(connectedProbe.altIpAddresses),
-			...connectedProbe.uuid !== uuid ? { uuid: connectedProbe.uuid } : {},
-		});
+	private async updateIds (currentAdoptedIp: string, connectedProbe: Probe) {
+		const update = async () => {
+			await this.sql(ADOPTED_PROBES_TABLE).where({ ip: currentAdoptedIp }).update({
+				ip: connectedProbe.ipAddress,
+				altIps: JSON.stringify(connectedProbe.altIpAddresses),
+				uuid: connectedProbe.uuid,
```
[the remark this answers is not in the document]
alexey-yarmosh: I don't think including current uuid of the probe into the update body was ever a problem. I tried that manually on staging and such updates were working fine. The real reason for `Duplicate entry 'zzz' for key 'gp_adopted_probes_uuid_unique'` is the same as for duplication of ip, just vice versa: old adoption tries to sync uuid by the ip, but that uuid is already occupied by the new adoption.

src/lib/override/adopted-probes.ts
alexey-yarmosh: I think it is possible that the adoption with earlier `date_created` syncs and becomes actual, and `date_updated` may change for many reasons (e.g. right now it is 21.11.2024 for all probes). As I see, we can remove `isCustomCity DESC` part and achieve the same behaviour.

MartinKolarik: OK.

src/lib/override/adopted-probes.ts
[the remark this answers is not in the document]
MartinKolarik: I can't replicate it manually either, but consider that #570 only changed the behavior when the UUID stayed the same and it did fix the issue. So:
 - If the UUID changed the update happens as before.
 - If the UUID didn't change, that means the probe already has that UUID in the DB. The UUID can't be in any other row since it's unique.

Maybe the catch could be in the part *If the UUID didn't change, that means the probe already has that UUID in the DB* because we're processing updates for 8 probes in parallel, and the probe objects are shared on the node side, so it's possible we did some changes in the memory which are not applied in the DB yet, but I can't think of the exact case that would lead to this error.

Anyway, let's keep the PR as it is I guess, but extend this log at least to log alt IPs and UUIDs too - in case we find there's still some problem, it'll be easier to understand:

```
logger.warn('Deleting probes with ids:', excessProbesIds);
```

Issue #576, opened 2024-11-22
Filter incoming alt IPs


alexey-yarmosh:
- ignore anycast alt IPs;
- ignore if alt IP is in different country than primary IP.

PR #581, opened 2024-11-28
feat: filter incoming alt IPs

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/576

Notes:
- During deploys number of geoip requests will double (we have 1300 primary ips + ~1300 alt ips). Need to check our limits.

### DIFF
--- a/probes-stats/known.ts
+++ b/probes-stats/known.ts
@@ -7,13 +7,13 @@
 
 import fs from 'node:fs';
 import { getRedisClient, initRedisClient } from '../src/lib/redis/client.js';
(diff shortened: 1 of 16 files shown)


jimaek:
How is it possible that we have so many alt IPs? We don't store private IPs as alt, right?

alexey-yarmosh:
There are a few probes with many (up to 30) IPv6 IPs. If there is a rule by which we can filter them, we can add it too.

MartinKolarik (approved):

PR #585, opened 2024-12-12
feat: prevent IP / alt IP conflicts

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping-dash-directus/issues/32.

Additionally:
- Previously first call of `syncDashboardData` was fired before any probe is connected to the worker, which triggered marking of all probes as `offline`. That is fixed.
- Fixed comparison of lat/long values, which triggered continuous updates of the probe.


MartinKolarik:
> Fixed comparison of lat/long values, which triggered continuous updates of the probe.

This was in case of custom cities, right?

alexey-yarmosh:
>This was in case of custom cities, right?

No, it depends on the value, e.g. SQL value `139.68` will be parsed by JS as `139.67999`.

Issue #586, opened 2024-12-12
Unstable e2e tests due to IPv6 docker daemon config


alexey-yarmosh:
Looks like [that](https://github.com/jsdelivr/globalping/commit/c7138c8ca98af25c097db08238b65de633dc4b02#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31e) makes the downloading of static files (malware ip, domains, etc.) sometimes break on CI.

Errors:
https://github.com/jsdelivr/globalping/actions/runs/12295409208/job/34312234491#step:8:647
https://github.com/jsdelivr/globalping/actions/runs/12283482017/job/34277304792#step:8:646

Here I commented that change out and download is fine again:
https://github.com/jsdelivr/globalping/commit/321da8747856f82b690ca022df38a1889c19aef0

MartinKolarik:
Fixed in f539cae85d7c791f218441fe0ba793d2e6755255.

MartinKolarik:
This is different from the original issue. We're running on a real IPv6 network now. Not sure why it failed here but it didn't happen to me yet.

MartinKolarik:
Closing as I haven't seen this for a while.

PR #587, opened 2024-12-13
fix: API build with IPv6 docker daemon config

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/586

Switching to `mode: client` didn't help, but looks like we can just run build step earlier.


MartinKolarik:
I already tried this 😄 it'll just fail in the Test E2E step where the docker images are built. Good news is we'll have a new CI service soon.

Issue #577, opened 2024-11-23
New APM and logging service


jimaek:
With New Relic getting pushy, and frankly its not very good either, we should start looking into alternatives. 

Some requirements:

- Low cost and predictable pricing
- Proper tracing with DB and Redis integration
- Server stats
- Logging without worrying
- Custom stats
- Simple to deploy and simple to use with good UX

jimaek:
Stuff to track for custom dashboards (all timeseries):

- Number of total measurements and per type
- Total number of measurements between users and anons(unique IP)
- Target hostnames to track abuse
- Registered users
- Total probes and adopted probes
- Probes per continent
- Stored measurements vs expired

MartinKolarik:
Closing as mostly done, ping me in DM for any minor adjustments. I don't really see a way to track "expired" measurements at the moment.

jimaek:
Doesn't Redis have that? Expired vs active keys

MartinKolarik:
Yes, but it's a single number as a total for the server. *Maybe* if we added our own tracking to capture the daily difference it would be somewhat useful.

jimaek:
Oh and lets build charts per reported client. So we would have charts for CLI, Slack, Web tools and separately all the other third-party (per UA)

PR #588, opened 2024-12-13
feat: switch to elastic APM (part of #577)

MartinKolarik:
Migrates the current setup, including the existing custom metrics. New metrics will be added separately.


package.json
```diff
@@ -117,14 +119,14 @@
...
 		"download:files": "tsx src/lib/download-files.ts",
 		"prepare": "npm run download:files; husky || echo 'Failed to install husky'",
-		"stats": "NEW_RELIC_ENABLED=false tsx probes-stats/known.ts",
-		"start": "node --max_old_space_size=3584 --max-semi-space-size=128 --experimental-loader newrelic/esm-loader.mjs -r newrelic dist/src/index.js",
-		"start:dev": "NODE_ENV=development FAKE_PROBE_IP=1 NEW_RELIC_ENABLED=false tsx src/index.ts",
-		"start:test": "NODE_ENV=test knex migrate:latest && NODE_ENV=test NEW_RELIC_ENABLED=false tsx src/index.ts",
+		"stats": "tsx probes-stats/known.ts",
+		"start": "ELASTIC_APM_CONFIG_FILE=elastic-apm-node.cjs node --max_old_space_size=3584 --max-semi-space-size=128 --experimental-loader elastic-apm-node/loader.mjs -r elastic-apm-node/start.js dist/src/index.js",
+		"start:dev": "NODE_ENV=development FAKE_PROBE_IP=1 ELASTIC_APM_CONFIG_FILE=elastic-apm-node.cjs tsx src/index.ts",
+		"start:test": "NODE_ENV=test knex migrate:latest && NODE_ENV=test ELASTIC_APM_CONFIG_FILE=elastic-apm-node.cjs tsx src/index.ts",
```
alexey-yarmosh, line 129: Is `ELASTIC_APM_CONFIG_FILE=elastic-apm-node.cjs` required here?

alexey-yarmosh:
For development purposes it may be handy for logger to accept inputs with more arguments (`logger.info('message', { a: 1 }, { a: 2 }, { a: 3 });`), or different type (`logger.info({ a: 1 });`). But if it is so strict to make all messages similar I think that is fine.

alexey-yarmosh (approved):

elastic-apm-node.cjs
```diff
@@ -0,0 +1,22 @@
...
+	version = fs.readFileSync(path.join(path.resolve(), 'data/LAST_API_COMMIT_HASH.txt'), 'utf8').trim();
+} catch {
+	version = require('./package.json').version;
+}
+
+module.exports = {
+	active: process.env.NODE_ENV === 'production',
+	serviceName: 'globalping-api',
+	serviceVersion: version,
+	logLevel: 'fatal',
```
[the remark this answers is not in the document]
MartinKolarik: It's a log level of the agent's own internal messages which aren't particularly useful, that's why we log only the highest level.

package.json
MartinKolarik: Yes because of the extension, it expects `js` otherwise.

MartinKolarik:
It's strict on purpose. The first argument being the message makes it easier to know where the error is from without looking down the stack trace.

The limited number of arguments is to encourage doing e. g. `log(message, error, { req, res })` instead of `log(message, error, req, res)` so that it's clear what each value represents (because key names will be in the output).

It also enables customizing output for [specific keys](https://github.com/MartinKolarik/h-logger2/blob/13ba2beda491ad109c5dcae1a154277aef03ec59/src/Logger.js#L95) so when you pass in a request, for example, it only logs a few useful properties instead of hundreds of lines.

alexey-yarmosh:
Ok, that makes sense 👍

MartinKolarik:
I think we can move the datetime, which will make it much better. The level should be near the message and can have a different number of letters as well.

PR #594, opened 2024-12-18
fix: adopted probes list should consider admin location overrides

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/592

### DIFF
--- a/src/lib/server.ts
+++ b/src/lib/server.ts
@@ -26,10 +26,11 @@ export const createServer = async (): Promise<Server> => {
 	await populateIpWhiteList();
 	// Populate cities info
 	await populateCitiesList();
+	// Populate Dashboard override data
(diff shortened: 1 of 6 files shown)


src/lib/server.ts
```diff
@@ -26,10 +26,11 @@ export const createServer = async (): Promise<Server> => {
 	await populateIpWhiteList();
 	// Populate cities info
 	await populateCitiesList();
+	// Populate Dashboard override data
+	await probeOverride.fetchDashboardData();
```
alexey-yarmosh, line 30: That change is made to fetch override data before using it during `initWsServer()`:
https://github.com/jsdelivr/globalping/blob/master/src/lib/ws/synced-probe-list.ts#L185-L186

MartinKolarik (approved):

Issue #541, opened 2024-08-28
Mark user owned probes in API responses


jimaek:
If the user is authenticated when making a new measurement we should clearly mark the probes he owns in the response.
Probably a system tag? This way the user could query his own probes directly as well.

Once the API supports it we can add markers to all of our tools.

Context https://github.com/jsdelivr/globalping-cli/issues/128

MartinKolarik:
We discussed two options here.

1) The tag is something like `my-probes` - anonymous, not directly tied to the user account; it would only mean that the user who started the measurement owns the probe.
2) The tag is based on the username, e.g., `u-jimaek`. This means other people would be able to use these tags as well, and the probes wouldn't be anonymous. It would be opt-in/out.

I like the second option more because it solves a more generic problem. E.g. jsdelivr/www.jsdelivr.com#689 could use the same tags then.

jimaek:
I dont think it applies to https://github.com/jsdelivr/www.jsdelivr.com/issues/689, the original text is outdated, the idea is to have a page per user, not per probe. So no opt-in per probe, its all or nothing, an account level setting.

So it sounds like the first option is better since option 2 only brings more issues

MartinKolarik:
Option two assumes one account level setting as well. Basically, if you enable the page, we might as well tag all the probes (since they are listed under your name on that page). And at that point I don't think we need to add anything here.

jimaek:
So a measurement response from any user or IP would always include the username in a tag and the tools would have to compare it to logged in user to highlight in the UI?

I guess that works. But it forces users to take extra steps and enable a feature they might not care about to simply get a small but useful UI feature.

It sounds like option 1 is needed for the UI highlighting and option 2 is a separate feature of public pages that makes your probes requestable by username

MartinKolarik:
We need to make the toggle for public pages as easy to find as possible anyway to get people to enable the probes pages. Most likely, we'll be prompting them right after adopting a probe. So not really extra steps. Also:

> But it forces users to take extra steps and enable a feature they might not care about to simply get a small but useful UI feature.

The UI highlighting is very minor and mostly useless itself. I don't think it deserves its own mechanism when it isn't needed and you can even look at this as a benefit - if people actually want it, we'll get more probe pages too.

jimaek:
The UI highlighting is what actually started this conversation and what people asked for. So its definitely not useless.

[...]

PR #607, opened 2025-01-22
feat: mark user owned probes in API responses

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/541

### DIFF
--- a/test/tests/unit/override/adopted-probes.test.ts
+++ b/test/tests/unit/override/adopted-probes.test.ts
@@ -767,7 +767,6 @@ describe('AdoptedProbes', () => {
 	});
 
 	it('getUpdatedLocation method should return updated location', async () => {
-		delete process.env['SHOULD_SYNC_ADOPTIONS'];
 		const adoptedProbes = new AdoptedProbes(sqlStub, getProbesWithAdminData);
 		sql.select.resolves([{
 			...defaultAdoptedProbe,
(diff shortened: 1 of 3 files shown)


test/tests/unit/override/adopted-probes.test.ts
```diff
@@ -412,16 +416,14 @@ describe('AdoptedProbes', () => {
...
-			subject: `Your probe's location has changed`,
-			message: 'Globalping detected that your [probe with IP address **1.1.1.1**](/probes/p-1) has changed its location from Ireland to United Kingdom. The custom city value "Dublin" is not applied anymore.\n\nIf this change is not right, please report it in [this issue](https://github.com/jsdelivr/globalping/issues/268).',
+			subject: 'Your probe\'s location has changed',
 		});
 
-		expect(sql.raw.args[1]![1]).to.deep.equal({
+		expect(sql.raw.args[1]![1]).to.deep.include({
 			recipient: '3cff97ae-4a0a-4f34-9f1a-155e6def0a45',
 			subject: `Your probe's location has changed`,
-			message: 'Globalping detected that your probe [**probe-gb-london-01**](/probes/p-9) with IP address **9.9.9.9** has changed its location from Ireland to United Kingdom. The custom city value "Dublin" is not applied anymore.\n\nIf this change is not right, please report it in [this issue](https://github.com/jsdelivr/globalping/issues/268).',
```
MartinKolarik, line 424: The formatting here is rather tricky so having a test for it doesn't seem like a bad idea, why remove it?

alexey-yarmosh: I though it is excess work to fix tests on every message update. And since it is .md format, update will still require manual check.

MartinKolarik: It's some extra work, but it's very easy to break the message, especially since there are multiple variants of it. That's why each variant got included in the tests as well.

alexey-yarmosh: Ok, I'll get it back.

MartinKolarik (approved):

Issue #600, opened 2025-01-18
Test results should be immutable if not in-progress


MartinKolarik:
I've run into a weird case where a test was marked as a timeout, but a second later, the timeout message was overwritten with the results. Setting aside the fact that the probe shouldn't have taken 32 seconds to report the results, the API should still prevent this as it is confusing and contradicts our documentation.

We already have a [check](https://github.com/jsdelivr/globalping/blob/66a4b489ae45254ee734486067633f802c916949/src/lib/redis/scripts.ts#L37) here, but when a test is marked as a timeout, we don't seem to increment `probesAwaiting`. Additionally, there's no check when calling `recordProgress()`, I think we should use a lua script with a check there as well.

PR #611, opened 2025-01-29
feat: immutable test results if not in-progress

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/600

### DIFF
--- a/src/lib/redis/scripts.ts
+++ b/src/lib/redis/scripts.ts
@@ -5,7 +5,7 @@ type RecordProgressScript = {
 	NUMBER_OF_KEYS: number;
 	SCRIPT: string;
 	transformArguments (measurementId: string, testId: string, keyToValue: TestProgress | HttpProgress): string[];
-	transformReply (reply: string): null;
+	transformReply (): null;
 } & {
 	SHA1: string;
 };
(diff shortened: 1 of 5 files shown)


src/measurement/store.ts
```diff
@@ -145,6 +136,7 @@ export class MeasurementStore {
 
 		await Promise.all([
 			this.redis.hDel('gp:in-progress', ids),
+			ids.map(id => this.redis.del((getMeasurementKey(id, 'probes_awaiting')))),
```
MartinKolarik, line 139: This should be `...` like the line below, no?

alexey-yarmosh: Indeed, fixed.

src/lib/redis/scripts.ts
```diff
@@ -20,10 +38,90 @@ type MarkFinishedScript = {
...
+	redis.call('JSON.SET', keyMeasurementResults, '$.updatedAt', date)
+	`,
+	transformArguments (measurementId, testId, keyToValue) {
+		return [
+			// keys
+			`gp:m:{${measurementId}}:results`,
+			`gp:m:{${measurementId}}:probes_awaiting`,
+			// values
+			testId,
+			JSON.stringify(Object.fromEntries(Object.entries(keyToValue).map(([ key, value ]) => [ `$.results[${testId}].result.${key}`, JSON.stringify(value) ]))),
```
MartinKolarik, line 116: Considering the update always has only 1 or 2 keys, maybe it would be better to pass each of them separately and have a simple `if` in the Lua script to avoid the JSON reencoding.

alexey-yarmosh: HTTP type update may have 3 key, `rawHeaders`, `rawBody`, `rawOutput`. But I think the update still makes sense.

MartinKolarik (approved):

PR #614, opened 2025-01-31
feat: update message handling

alexey-yarmosh:
Fixes https://github.com/jsdelivr/meta/issues/51

Didn't add it as a part of lua script, as global hash structure is used. Plus checks are mostly performed in-memory, redis is a fallback.

### DIFF
--- a/src/lib/probe-validator.ts
+++ b/src/lib/probe-validator.ts
@@ -3,38 +3,39 @@ import TTLCache from '@isaacs/ttlcache';
 import { type RedisCluster, getMeasurementRedisClient } from './redis/measurement-client.js';
 
 export class ProbeValidator {
-	private readonly testIdToProbeId = new TTLCache<string, string>({
+	private readonly measurementIdToTests = new TTLCache<string, Map<string, string>>({
 		ttl: (config.get<number>('measurement.timeout') + 30) * 1000,
 	});
 
 	constructor (private readonly redis: RedisCluster) {}
 
 	addValidIds (measurementId: string, testId: string, probeUuid: string): void {
-		const key = ProbeValidator.getKey(measurementId, testId);
-		this.testIdToProbeId.set(key, probeUuid);
+		const measurement = this.measurementIdToTests.get(measurementId);
+
+		if (!measurement) {
+			this.measurementIdToTests.set(measurementId, new Map([ [ testId, probeUuid ] ]));
+		} else {
+			measurement.set(testId, probeUuid);
+		}
 	}
 
 	async validateProbe (measurementId: string, testId: string, probeUuid: string): Promise<void> {
-		const key = ProbeValidator.getKey(measurementId, testId);
-		let probeId = this.testIdToProbeId.get(key);
+		const measurement = this.measurementIdToTests.get(measurementId);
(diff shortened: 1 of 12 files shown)


src/lib/probe-validator.ts
```diff
@@ -0,0 +1,49 @@
...
+export class ProbeValidator {
+	private readonly testIdToProbeId = new TTLCache<string, string>({
+		ttl: (config.get<number>('measurement.timeout') + 30) * 1000,
+	});
+
+	constructor (private readonly redis: RedisCluster) {}
+
+	addValidIds (measurementId: string, testId: string, probeUuid: string): void {
+		const key = ProbeValidator.getKey(measurementId, testId);
+		this.testIdToProbeId.set(key, probeUuid);
```
MartinKolarik, line 14: I would instead use a regular `Map` stored at `measurementId`, with `testId` as its keys. Should, greatly reduce the amount of internal timers the TTLCache needs to deal with.

src/measurement/handler/request.ts
```diff
@@ -0,0 +1,12 @@
...
+import { getProbeValidator } from '../../lib/probe-validator.js';
+import { MeasurementRequestMessage } from '../types.js';
+
+export const listenMeasurementRequest = (probe: Probe) => (event: string, data: unknown) => {
+	if (event !== 'probe:measurement:request') {
+		return;
+	}
+
+	const message = data as MeasurementRequestMessage;
+	getProbeValidator().addValidIds(message.measurementId, message.testId, probe.uuid);
```
MartinKolarik, line 11: Any reason we're adding it indirectly here instead of inside `sendToProbes` ?

src/measurement/store.ts
```diff
@@ -73,14 +71,17 @@ export class MeasurementStore {
 			results,
 		};
 		const measurementWithoutDefaults = this.removeDefaults(measurement, request);
+		const testsToProbes = Object.fromEntries(Array.from(onlineProbesMap, ([ testId, probe ]) => [ `${id}_${testId}`, probe.uuid ]));
 
 		await Promise.all([
 			this.redis.hSet('gp:in-progress', id, startTime.getTime()),
-			this.redis.set(getMeasurementKey(id, 'probes_awaiting'), onlineProbesMap.size, { EX: probesAwaitingTtl }),
+			this.redis.set(getMeasurementKey(id, 'probes_awaiting'), onlineProbesMap.size, { EX: config.get<number>('measurement.timeout') + 30 }),
```
MartinKolarik, line 78: Any particular reason we're increasing this as well?

src/measurement/handler/request.ts
alexey-yarmosh: If called inside `sendToProbes` it will add ids to the process initiating the measurement. While what we need is to add id to the process where particular probe is connected, so every process listens for `probe:measurement:request` event using `socket.onAnyOutgoing` and stores the id in memory before sending message to the probe.

src/measurement/store.ts
alexey-yarmosh: Just for the value to be the same, as there is no reason for delay to be `+5` in one case and `+30` in another.

PR #630, opened 2025-03-10
fix: add Great Britain as GB alias

MartinKolarik:
United Kingdom is the official name, so we already have that in the index. On the other hand, we were missing Great Britain.

### DIFF
--- a/src/lib/location/countries.ts
+++ b/src/lib/location/countries.ts
@@ -5,7 +5,7 @@
  * ]
  * */
 export const aliases = [
-	[ 'gb', 'uk', 'united kingdom', 'england', 'northern ireland', 'wales', 'scotland' ],
+	[ 'gb', 'uk', 'great britain', 'england', 'northern ireland', 'wales', 'scotland' ],
 ];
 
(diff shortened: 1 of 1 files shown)


MartinKolarik:
The IP we used in e2e tests changed its location, so the tests started to fail. I changed it to use `fakeLookup()` in e2e instead.

src/lib/location/countries.ts
```diff
@@ -5,7 +5,7 @@
  * ]
  * */
 export const aliases = [
-	[ 'gb', 'uk', 'united kingdom', 'england', 'northern ireland', 'wales', 'scotland' ],
+	[ 'gb', 'uk', 'great britain', 'england', 'northern ireland', 'wales', 'scotland' ],
```
alexey-yarmosh, line 8: Shouldn't `united kingdom` still be in the list?

MartinKolarik: > United Kingdom is the official name, so we already have that in the index.

alexey-yarmosh (approved):

PR #645, opened 2025-04-10
feat: add token support to demo page

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/531


MartinKolarik (commented):
I was thinking an input field here, nobody would discover or remember this param.

MartinKolarik (approved):

Issue #672, opened 2025-06-21
Timed out measurements stuck in 'in-progress'


alexey-yarmosh:
Most of the timed out measurements are not marked as `finised` and are stuck in `in-progress` state. E.g.:
```json
{
  "id": "HUOMERVBnomIe6Oq",
  "type": "http",
  "status": "in-progress",
  "createdAt": "2025-06-21T01:12:07.978Z",
  "updatedAt": "2025-06-21T01:12:12.419Z",
  "target": "www.333brazil.com",
  "limit": 60,
  "probesCount": 60,
  "locations": [
    {
      "magic": "Brazil"
    }
  ],
  "measurementOptions": {
    "port": 443,
    "request": {
      "method": "GET"
    }
  },
  "results": [...]
}
```
That happens because of that `mGet` in cluster mode. It returns only values from the same hash slots, and `null`s for all others:
https://github.com/jsdelivr/globalping/blob/d2fe57fcf1788c60142e4280451704c0a6ae7727/src/measurement/store.ts#L121

PR #676, opened 2025-06-27
fix: timed out measurements stuck in "in-progress"

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/672

### DIFF
--- a/src/measurement/store.ts
+++ b/src/measurement/store.ts
@@ -141,12 +141,13 @@ export class MeasurementStore {
 			}
 		}
 
-		const updateMeasurementPromises = measurements.map(measurement => this.redis.json.set(getMeasurementKey(measurement.id), '$', measurement));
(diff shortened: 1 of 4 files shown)


src/measurement/store.ts
```diff
@@ -140,7 +141,7 @@ export class MeasurementStore {
 			}
 		}
 
-		const updateMeasurementPromises = existingMeasurements.map(measurement => this.redis.json.set(getMeasurementKey(measurement.id), '$', measurement));
+		const updateMeasurementPromises = measurements.map(measurement => this.redis.json.set(getMeasurementKey(measurement.id), '$', measurement));
```
MartinKolarik: > We should also improve the writing part below and use Bluebird.map() with concurrency no more than 100 instead of creating up to 10k queries at once, like now.

MartinKolarik (approved):

PR #688, opened 2025-08-03
fix: tests limit for the authenticated users

alexey-yarmosh:


src/measurement/schema/global-schema.ts
```diff
@@ -15,7 +15,7 @@ export const schema = Joi.object({
 	target: targetSchema,
 	measurementOptions: measurementSchema,
 	locations: locationSchema,
-	limit: Joi.number().min(1).when('$userId', {
+	limit: Joi.number().min(1).when('$user', {
```
MartinKolarik: This is not correct either, as it gives higher limits to anonymous tokens. We need to check for `user.id`.

jimaek: Im now curious, is there anything we could have done to make Coderabbit catch this issue automatically?

MartinKolarik: Not really, the Coderabbit reviews are very random. It misses stuff that matters all the time, and the comments it makes are wrong half the time as well.

MartinKolarik: I'd say the main reason is the lack of context. It only reads the diffs, so when the source of the name (or behavior, in general) is outside of the PR, it can't catch the inconsistency.

Ernst2020:
Excellent that this “bug” was found an swatted. Going from well over a thousand measurements a day to single digit numbers made me wonder whether keeping my 7 probes live was worthwhile.

Will probes update themselves automatically?

MartinKolarik (approved):

Issue #383, opened 2023-06-30
Magic field wildcard support


jimaek:
How about we add wildcard support using * ?

It will be useful in cases where there is full match but you dont want it. e.g. There are Zenlayer probes and Zenlayer Inc. probes.
magic:zenlayer will only return 3 probes because it matches them fully.

A good solution would be magic:zenlayer* to match every single probe. Other companies have similar issues because their ASNs are not uniform and use random combinations of their company names.

MartinKolarik:
It seems a little confusing because there's an implicit `*` already if there's no exact match. Maybe we could instead try to normalize the ASNs?

jimaek:
>It seems a little confusing because there's an implicit * already if there's no exact match

Internally yes but it's visible to the user.

>Maybe we could instead try to normalize the ASNs?

They are set by the providers, not sure how we could normalize them. It could be anything

MartinKolarik:
Possible starting point: https://github.com/ProfoundNetworks/company_designator/blob/master/company_designator.yml (the `abbr` fields)

MartinKolarik:
One remaining problem that I see after checking our data is `Vodafone`, which has various national entities, e.g., `Vodafone Portugal` and `Vodafone Germany`. There I don't see a better solution right now... But regardless, I think we should aim to fix most of the cases by normalizing the data and not expecting the users to use different query forms.

MartinKolarik:
I came across this today and had a new idea regarding the networks. We can change the logic so that, in case of networks, a "full match" is based on words, not the whole value, so:

1. For query "Vodafone", networks "Vodafone", "Vodafone Portugal", or "Vodafone LLC" are all considered *exact* matches.
2. For query "Voda", they would all be considered *partial* matches, like now, and excluded if there was an actual provider with that exact name.

I still think the name normalization is useful too, as it simplifies presentation (e.g., 1 value in the location search instead of 4), but it doesn't solve the issue with national entities on its own.

MartinKolarik:
Preview of the suffix normalization based on https://www.gleif.org/en/lei-data/code-lists/iso-20275-entity-legal-forms-code-list, seems very promising: https://1drv.ms/x/c/df6526c9b409d48f/EWzJQ5TBJqZHjm2PTLs7euIBMMH9zTTM2i6f1Uj-3UB8Zg?e=mIhZKc

PR #720, opened 2025-09-16
feat: normalize network names

MartinKolarik:
Implements the normalization part from https://github.com/jsdelivr/globalping/issues/383#issuecomment-1618206231.

The network names are now much more consistent, which fixes lots of issues in the location autocomplete, looking glass pages, etc. The approach uses lots of regex magic based on iterative tests, but produces very good results.


src/lib/geoip/legal-name-normalization.ts
```diff
@@ -0,0 +1,175 @@
...
+		// Remove one or more patterns from the end; multiple patterns are only removed if:
+		//  - the first patterns ends with a dot or a closing parenthesis, optionally followed by a comma, or,
+		//  - the first pattern is followed by an ampersand or a hyphen.
+		return new RegExp(`\\s+(?:(?:${patterns.join('|')})(?:(?<=[.)]),?\\s*|\\s*[&-]\\s*|$))+$`, 'i');
+	};
+
+	return {
+		namesPattern: buildPattern(preparedNames),
+		abbrsPattern: buildPattern(preparedAbbrs),
+	};
```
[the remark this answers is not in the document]
MartinKolarik: Not an issue as this is not (directly) user-controlled.

src/lib/geoip/legal-name-normalization.ts
```diff
@@ -0,0 +1,175 @@
...
+			.replace(/(?<=\w)(?=\w)/g, '[.\\s]*')
+			// Make the final dot optional.
+			.replace(/\.$/, '.?');
+
+		// Allow the whole abbreviation to be wrapped in parentheses.
+		preparedAbbrs.push(`\\(?${abbrPattern}\\)?`);
+	}
+
+	const buildPattern = (patterns: string[]) => {
+		// Remove one or more patterns from the end; multiple patterns are only removed if:
```
alexey-yarmosh: Maybe we should also search at the beginning? We have e.g. "LLC Baxet".

src/lib/geoip/legal-name-normalization.ts
```diff
@@ -0,0 +1,175 @@
...
+	if (!namesPattern || !abbrsPattern) {
+		throw new Error('Legal name normalization is not initialized.');
+	}
+
+	return name.trim()
+		// Normalize "trading as" names, e.g., "Matteo Martelloni trading as DELUXHOST" => "DELUXHOST"
+		.split(/\s+trading as\s+/i).at(-1)!
+		// Apply the main cleanup patterns.
+		.replace(namesPattern, '').trim()
+		.replace(abbrsPattern, '').trim()
```
alexey-yarmosh: Looks like regex search might take a lot of time: https://github.com/jsdelivr/globalping/actions/runs/17804117023/job/50611441251#step:7:45 
It takes even more localy.

Maybe we can add more variants to `legalFormsLong` and `legalFormsAbbr`? E.g. `llc`, `l.l.c.` and `l l c`. Then split the network value by spaces, commas, parens, etc. and use `has()` for every part starting from the end?

src/lib/geoip/legal-name-normalization.ts
MartinKolarik: I noticed these, but the cases are relatively rare, and the more flexibility we add, the higher the chances of false-positive matches.

src/lib/geoip/legal-name-normalization.ts
MartinKolarik: The first two calls are slow, then it gets optimized and takes about 20ms per call. I considered caching the results, too, but after the two calls, it should be good enough since it happens only on the probe connection (but for every geoip provider, so maybe I will add the caching, actually).

MartinKolarik: > Then split the network value by spaces, commas, parens, etc. and use has() for every part starting from the end?

This would be much slower with the high number of variants. The regexes are generaly the fastest way to check for many possible values in V8.

src/lib/geoip/legal-name-normalization.ts
MartinKolarik: Fixed by allowing prefixes for specific countries that use them.

src/lib/geoip/legal-name-normalization.ts
```diff
@@ -0,0 +1,229 @@
...
+		preparedAbbrs.push(`\\(?${abbrPattern}\\)?`);
+	}
+
+	return {
+		namesPattern: builder(preparedNames),
+		abbrsPattern: builder(preparedAbbrs),
+	};
+}
+
+async function collectLegalForms (legalFormsData: CsvLegalFormRow[], synthesizeAbbreviations: number = 2) {
```
MartinKolarik: Note for myself: rename the param.

PR #722, opened 2025-09-18
feat: normalize regional network names

### LINKED ISSUE #721: Normalize national network names
After #720, `xTom Pty Ltd` and `xTom GmbH` will become just `xTom`, but `xTom Japan Corporation` will become `xTom Japan`. The same happens with a number of other providers (Vodafone, Orange, ...).

Thanks to #718, this no longer causes issues in matching, but it still means any client working with the probe list will see these as multiple separate networks unless it does its own normalization. E.g., our autocomplete at https://globalping.io/ will suggest them separately, and (in case of xTom) we'll still have two separate looking glass pages (instead of three).

I'm thinking it might make sense to use a similar approach to #720 and collapse all the national entities into one. It should be relatively easy and would make working with the probe list generally easier. A small potential downside is that it would no longer be possible to target a specific national network like `xTom Japan`, but since we support combined filters, the same can be trivially achieved with `xTom+Japan` (this also has the advantage of working in all cases, even if the network doesn't have separate national entities).

MartinKolarik:
Closes #721.

To avoid false positives, the approach is based on two data files that live directly in the repo. Generating them was rather complex, but we likely won't need to make big changes to them in the future (and small changes are easy to do by hand).

The data files were created in roughly these steps:
   1. Take unique ASN names from [ASN info](https://github.com/ipverse/asn-info) (around 90k records).
   2. Strip legal suffixes using the existing code in #720.
   3. Group the ASNs by the first word in their name. Drop groups that only have one member.
   4. In each group, compute its word-based LCS, and add the remaining words of each name into a "possible suffixes" file.
   5. Let an AI review the "possible suffixes" file to get regional suffixes from it. Additionally, ask it to add likely words based on `countries-list`. Store the results as "regional suffixes".
   6. Take a flat list of all the ASNs from step 4), and strip the regional suffixes based on the existing list.
   7. Generate a diff of the changes (around 1.5k records), and let AI classify each change as "correct", "wrong", or "not sure". Follow by manual review.

I committed both "correct" and "wrong" files from step 7) in case we want to review them later, even though the "wrong" is not used by the app.

### DIFF
--- a/src/lib/geoip/legal-name-normalization.ts
+++ b/src/lib/geoip/legal-name-normalization.ts
@@ -145,7 +145,7 @@ export const normalizeLegalName = (name: string) => {
 	const international = normalized.replace(nationalSuffixesPattern, '').trim();
 
 	// Use the result only if it's on our list of verified names.
-	if (international && internationalProviders.has(ascii(international.toLowerCase()))) {
+	if (international && internationalProviders.has(international.toLowerCase())) {
 		normalized = international;
 	}
 
--- a/src/lib/geoip/utils.ts
+++ b/src/lib/geoip/utils.ts
@@ -13,7 +13,9 @@ export const normalizeCityName = (name: string): string => normalizeCityNamePubl
 
 export const normalizeFromPublicName = (name: string): string => name.toLowerCase();
 
-export const normalizeNetworkNamePublic = (name: string): string => normalizeLegalName(name);
+export const normalizeNetworkNamePublic = (name: string): string => {
+	return normalizeLegalName(anyAscii(name));
+};
 
 export const normalizeNetworkName = (name: string): string => normalizeNetworkNamePublic(name).toLowerCase();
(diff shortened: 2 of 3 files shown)


src/lib/geoip/legal-name-normalization.ts
```diff
@@ -146,6 +164,15 @@ function buildPatterns (names: string[], abbrs: string[], builder: (patterns: st
 	};
 }
 
+function buildNationalSuffixesPattern (words: string[]) {
+	const preparedWords = words
+		.map(w => ascii(w))
+		.sort((a, b) => b.length - a.length)
+		.map(w => _.escapeRegExp(w));
+
+	return new RegExp(`\\s\\(?(?:${preparedWords.join('|')})\\)?$`, 'i');
```
MartinKolarik: One more regex but this one is fast 😄

Issue #684, opened 2025-07-16
Improve alt IP reporting flow


MartinKolarik:
Related to https://github.com/jsdelivr/globalping/issues/649#issuecomment-3078385590. The current flow works fine for reporting "it works" case, but wouldn't work for reporting "it no longer works" case (if a request times out and the IP was previously registered, it won't be removed).

Here's an idea for an alternative approach that fixes this and also seems more reliable than what we do now:
1. A probe makes a POST request to our API endpoint. There is no token in the request. Instead, there's a token + IP in the response. The API *doesn't* add the alt IP at this step, it simply stores the token + IP pair in redis.
2. The probe then reports via its existing websocket connection that either a) the IP worked, and sends the token to prove it b) the IP didn't work.
3. The API adds/removes the IP based on the WebSocket message.

This way, we also don't need any "synced list remote updates" because the writes are always done by the node that owns the probe, and step 1 doesn't need probe info at all.

PR #731, opened 2025-09-25
feat: improve alt IP reporting flow

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/684
part of https://github.com/jsdelivr/globalping-probe/issues/310

### DIFF
--- a/src/alternative-ip/route/alternative-ip.ts
+++ b/src/alternative-ip/route/alternative-ip.ts
@@ -4,12 +4,12 @@ import { RateLimiterMemory, RateLimiterRes } from 'rate-limiter-flexible';
 import { getAltIpsClient } from '../../lib/alt-ips-client.js';
 import type { ExtendedContext } from '../../types.js';
 
-export const rateLimiter = new RateLimiterMemory({
+const rateLimiter = new RateLimiterMemory({
 	points: 20,
 	duration: 60,
 });
 
-export const checkRateLimit = async (ctx: ExtendedContext) => {
+const checkRateLimit = async (ctx: ExtendedContext) => {
 	const ip = ctx.request.ip;
 
 	try {
--- a/src/adoption/adoption-token.ts
+++ b/src/adoption/adoption-token.ts
@@ -111,7 +111,7 @@ export class AdoptionToken {
 		let dProbe: DProbe | null = this.adoptedProbes.getByIp(probe.ipAddress) || this.adoptedProbes.getByUuid(probe.uuid);
 
 		if (!dProbe || dProbe.userId !== user.id) {
-			dProbe = await this.fetchProbe(probe);
+			dProbe = await this.fetchDProbe(probe);
 		}
 
 		if (dProbe && dProbe.userId === user.id) {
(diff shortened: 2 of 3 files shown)


src/alternative-ip/route/alternative-ip.ts
```diff
@@ -1,31 +1,41 @@
...
-const handle = async (ctx: Context): Promise<void> => {
-	const request = ctx.request.body as AlternativeIpRequest;
+import { RateLimiterMemory, RateLimiterRes } from 'rate-limiter-flexible';
+import { getAltIpsClient } from '../../lib/alt-ips-client.js';
+import type { ExtendedContext } from '../../types.js';
+
+export const rateLimiter = new RateLimiterMemory({
+	points: 100,
+	duration: 60,
+});
```
alexey-yarmosh, line 10: I think we need a rate limiter on this endpoint. But I am not sure about its configuration:
1. Rate limiting here shouldn't be precise, so using memory instead of writing to redis looks good.
2. Duration is set to 60 because redis TTL is 60.
3. Points should be calculated ~like `numberOfPossibleIpDuplicates * numberOrPossibleReconnectionsIn60Seconds`.

MartinKolarik: If by "duplicates" you mean the same probe would have the same outgoing IP on multiple interfaces, we could include the resolved IP in the error response and let the probe handle it gracefully (if the same IP already registered, any other errors are ignored).

alexey-yarmosh: Yes, I mean that. API treats such requests as identical so they will hit rate limit faster.

MartinKolarik: Ok then handle on the probe side and change the limit here to just slightly above `numberOrPossibleReconnectionsIn60Seconds`... 20 maybe?

alexey-yarmosh: Updated the `points`. Probe side update is done [here](https://github.com/jsdelivr/globalping-probe/pull/320/commits/10379fcd39478554481910d3e600785abdefe076).

src/alternative-ip/route/alternative-ip.ts
```diff
@@ -1,31 +1,41 @@
...
+import { RateLimiterMemory, RateLimiterRes } from 'rate-limiter-flexible';
+import { getAltIpsClient } from '../../lib/alt-ips-client.js';
+import type { ExtendedContext } from '../../types.js';
+
+export const rateLimiter = new RateLimiterMemory({
+	points: 20,
+	duration: 60,
+});
+
+export const checkRateLimit = async (ctx: ExtendedContext) => {
```
MartinKolarik: Looks like we don't need to export this.

src/probe/handler/alt-ips.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const handleAltIps = (probe: Probe) => async (ipsToTokens: [string, string][], callback: (result: { addedAltIps: string[]; rejectedIpsToReasons: Record<string, string> }) => void) => {
+	const validation = altIpsSchema.validate(ipsToTokens);
+
+	if (validation.error) {
+		throw validation.error;
+	}
+
+	const altIpsClient = getAltIpsClient();
+	const { addedAltIps, rejectedIpsToReasons } = await altIpsClient.addAltIps(probe, validation.value);
```
MartinKolarik, line 13: Type error here.

PR #737, opened 2025-10-01
refactor: replace `Probe` type with `ServerProbe` and `SocketProbe`

MartinKolarik:
Splits the `Probe` type into:
 - `SocketProbe` - which represents the object that lives on the instance that handles the connection, serves as the data source in the synced probe list, and does not have the location overrides applied,
 - `ServerProbe` - which represents read-only information about the probe shared across servers, with the location overrides applied.

The goal is to prevent bugs where we use one instead of the other. Fixes a bug in `addAltIps()` where the location comparison was done on the raw data, not on the user-corrected location.

### DIFF
--- a/src/probe/schema/probe-response-schema.ts
+++ b/src/probe/schema/probe-response-schema.ts
@@ -1,8 +1,8 @@
 import Joi from 'joi';
-import { ServerProbe, ProbeStats } from '../types.js';
+import { SocketProbe, ProbeStats } from '../types.js';
 import { globalIpOptions } from '../../measurement/schema/utils.js';
 
-export const statusSchema = Joi.string<ServerProbe['status']>().valid('initializing', 'ready', 'unbuffer-missing', 'ping-test-failed', 'sigterm').required();
+export const statusSchema = Joi.string<SocketProbe['status']>().valid('initializing', 'ready', 'unbuffer-missing', 'ping-test-failed', 'sigterm').required();
 
 export const ipVersionSchema = Joi.boolean().required();
 
--- a/src/probe/schema/probe-response-schema.ts
+++ b/src/probe/schema/probe-response-schema.ts
@@ -1,8 +1,8 @@
 import Joi from 'joi';
-import { Probe, ProbeStats } from '../types.js';
+import { ServerProbe, ProbeStats } from '../types.js';
 import { globalIpOptions } from '../../measurement/schema/utils.js';
 
-export const statusSchema = Joi.string<Probe['status']>().valid('initializing', 'ready', 'unbuffer-missing', 'ping-test-failed', 'sigterm').required();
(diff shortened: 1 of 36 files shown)


src/probe/schema/probe-response-schema.ts
```diff
@@ -1,8 +1,8 @@
 import Joi from 'joi';
-import { Probe, ProbeStats } from '../types.js';
+import { ServerProbe, ProbeStats } from '../types.js';
 import { globalIpOptions } from '../../measurement/schema/utils.js';
 
-export const statusSchema = Joi.string<Probe['status']>().valid('initializing', 'ready', 'unbuffer-missing', 'ping-test-failed', 'sigterm').required();
+export const statusSchema = Joi.string<ServerProbe['status']>().valid('initializing', 'ready', 'unbuffer-missing', 'ping-test-failed', 'sigterm').required();
```
alexey-yarmosh: Looks like here `SocketProbe` should be used, since handler uses it.

test/tests/unit/override/admin-data.test.ts
```diff
@@ -63,8 +63,8 @@ describe('AdminData', () => {
 
 		await adminData.syncDashboardData();
 		const updatedProbes = adminData.getUpdatedProbes([
-			{ ipAddress: '2.2.2.2', altIpAddresses: [ '1.1.1.1' ], location: { city: 'Miami', state: 'FL' } } as unknown as Probe,
-			{ ipAddress: '3.3.3.3', altIpAddresses: [], location: { city: 'Miami', state: 'FL' } } as unknown as Probe,
+			{ ipAddress: '2.2.2.2', altIpAddresses: [ '1.1.1.1' ], location: { city: 'Miami', state: 'FL' } } as unknown as ServerProbe,
```
alexey-yarmosh: `getUpdatedProbes` returns `SocketProbe[]` as I see.

test/tests/unit/override/adopted-probes.test.ts
```diff
@@ -394,7 +394,7 @@ describe('AdoptedProbes', () => {
 					network: 'The Constant Company',
 					allowedCountries: [ 'GB' ],
 				},
-			} as Probe,
+			} as ServerProbe,
```
alexey-yarmosh: Same.  `getUpdatedProbes` returns `SocketProbe[]` as I see.

alexey-yarmosh (changes requested):

test/tests/unit/override/adopted-probes.test.ts
MartinKolarik: I actually don't see why we cast it here in the first place. Removed.

alexey-yarmosh (approved):

PR #743, opened 2025-10-03
feat: add global cloud tags

### LINKED ISSUE #739: Support cloud providers in `magic` and `network`
While `magic: "gcp"` location is working, `magic: "aws"` does not. 
Also `network: "gcp"` and `network: "aws"` are not working too.

I think we should also add "Azure" as network alias.

MartinKolarik:
Closes #739.

### DIFF
--- a/src/lib/alt-ips-client.ts
+++ b/src/lib/alt-ips-client.ts
@@ -8,7 +8,7 @@ import { ProbeError } from './probe-error.js';
 import { isIpBlocked } from './blocked-ip-ranges.js';
 import { getPersistentRedisClient, type RedisClient } from './redis/persistent-client.js';
 import { ServerProbe, SocketProbe } from '../probe/types.js';
-import { updateProbe } from '../probe/builder.js';
+import { updateProbeAltIps } from '../probe/builder.js';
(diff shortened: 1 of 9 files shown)


src/lib/alt-ips-client.ts
```diff
@@ -39,7 +39,7 @@ export class AltIpsClient {
 		const { ipsWithValidTokens, tokenErrors } = await this.validateTokens(ipsToTokens, serverProbe);
 		const { altIpAddresses, ipErrors } = await this.validateIps(ipsWithValidTokens, serverProbe);
 
-		updateProbe(probe, probe.ipAddress, altIpAddresses);
+		updateProbeAltIps(probe, altIpAddresses);
```
alexey-yarmosh, line 42: It might make sense to sort `altIpAddresses` before calling `updateProbeAltIps`. So array with multiple cloud alt IPs will always return the same cloud region. This should also result in less false results in `_.isEqual`.

MartinKolarik: https://github.com/jsdelivr/globalping/commit/46a94dab3175d154cd59c08b1a8642037ea2d3c9

PR #761, opened 2025-11-14
feat: add support for retrieving offloaded measurements

### LINKED ISSUE #598: Maria DB data offloading
As a follow-up to https://github.com/jsdelivr/globalping/issues/269#issuecomment-2564143624 we may want to offload data older than 1 hour from Redis to another storage so that we can keep the measurements results stored for longer. S3 and similar alternatives would be very expensive because of request based pricing, so Maria DB seems like the best option. All queries would be single-row lookups by primary key, so a server with 16 - 32 GB RAM and 2 TB+ of fast storage should be sufficient.

MartinKolarik:
Closes #598.

### DIFF
--- a/src/measurement/store.ts
+++ b/src/measurement/store.ts
@@ -15,7 +15,6 @@ import type {
 } from './types.js';
 import { getDefaults } from './schema/utils.js';
 import { getMeasurementRedisClient, type RedisCluster } from '../lib/redis/measurement-client.js';
-import { getPersistentRedisClient, type RedisClient } from '../lib/redis/persistent-client.js';
 import { AuthenticateStateUser } from '../lib/http/middleware/authenticate.js';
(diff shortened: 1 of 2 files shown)


src/measurement/store.ts
```diff
@@ -54,7 +61,27 @@ export class MeasurementStore {
 		this.offloader = new MeasurementStoreOffloader(measurementStoreClient, this);
 	}
 
-	async getMeasurementString (id: string): Promise<string> {
+	async getMeasurementString (id: string): Promise<string | null> {
+		try {
+			const { minutesSinceEpoch, userTier } = parseMeasurementId(id);
```
alexey-yarmosh: It can make sense to wrap only `const { minutesSinceEpoch, userTier } = parseMeasurementId(id);` in `try.. catch` to make expecting of error there explicit and remove nesting of two `try.. catch`es.

PR #773, opened 2025-12-22
test: add more http e2e tests

alexey-yarmosh:
Part of https://github.com/jsdelivr/globalping-probe/issues/148

### DIFF
--- a/test/e2e/cases/http.test.ts
+++ b/test/e2e/cases/http.test.ts
@@ -101,6 +101,26 @@ describe('http measurement', () => {
 		expect(response).to.matchApiSchema();
 	});
 
+	it('should have null tls and dns timings for HTTP request to IP', async () => {
+		const { id } = await got.post('http://localhost:80/v1/measurements', {
+			json: {
+				target: '2606:4700:3037::ac43:d071',
+				type: 'http',
+				measurementOptions: {
+					protocol: 'HTTP',
+				},
+			},
+		}).json<any>();
+
+		const response = await waitMeasurementFinish(id);
+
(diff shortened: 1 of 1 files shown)


test/e2e/cases/http.test.ts
```diff
@@ -144,4 +247,38 @@ describe('http measurement', () => {
...
+				},
+			},
+		}).json<any>();
+
+		const response = await waitMeasurementFinish(id);
+
+		expect(response.body.status).to.equal('finished');
+		expect(response.body.results[0].result.status).to.equal('failed');
+		expect(response).to.matchApiSchema();
+	});
```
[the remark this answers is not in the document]
alexey-yarmosh: Yes, lets try and if it brings problems - remove later.

.github/workflows/e2e.yml
```diff
@@ -63,5 +63,8 @@ jobs:
           npm ci
           npm run build
       - name: Test E2E
+        env:
+          CURRENT_BRANCH: ${{ github.head_ref || github.ref_name }}
         run: |
+          echo "Current branch: $CURRENT_BRANCH."
```
MartinKolarik: I believe we were not doing this intentionally to ensure we test backward compatibility with current probe versions as well. In this case, it means the tests fail until probe merge, but in other it might useful 🤔

alexey-yarmosh: We have this logic in probe repo, but now I see it makes sense to check newApi-oldProbe, but not oldApi-newProbe. So I'll remove this.

PR #805, opened 2026-03-03
feat: add too-many-disconnects status

alexey-yarmosh:
Part of https://github.com/jsdelivr/globalping-probe/issues/323


alexey-yarmosh:
Having a separate status will be helpful during debug and e.g. on the /demo page. Plus "ping test failed" is too specific to put reconnections there. To simplify the dash UI we can show "unstable network" for both.

MartinKolarik (approved):

PR #841, opened 2026-05-13
feat: credits notification

### LINKED ISSUE #198: Credits notification
> Running low on credits: either using some prediction based on the past usage, or when it goes below the configured amount (or an option to choose from both). This should also have an option to send emails after #27. 

 _Originally posted by @MartinKolarik in [#31](https://github.com/jsdelivr/globalping-dash/issues/31#issuecomment-2873906697)_

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping-dash/issues/198

Explicit threshold approach is used as the simplest to understand and use.


PavelKopecky (approved):

MartinKolarik (changes requested):
I don't think the API should be involved in this at all, especially as we might need to refactor the credits handling soon to introduce caching. Directus can do a simple `SELECT * FROM users join credits where credits < threshold` every few minutes, and we can add a flag on the user, e.g., `creditsBelowThreshold`, which decides whether a new notification is sent (not if already true). The flag resets on credit additions if the credits go above the threshold.

PR #849, opened 2026-06-05
feat: add more dc networks

alexey-yarmosh:
Adding overrides so OVH or Hetzner probes are not marked as eyeball.

Also Baxet uses both `Baxet` and `Baxet Group` network name. It might make sense to either merge them, or make regex broader `/baxet group/ -> /baxet/`.

### DIFF
--- a/src/lib/geoip/overrides.ts
+++ b/src/lib/geoip/overrides.ts
@@ -9,4 +9,14 @@ export const isHostingOverrides = [
 		normalizedNetwork: /psychz networks/,
 		isHosting: true,
 	},
+	// https://www.hetzner.com/
+	{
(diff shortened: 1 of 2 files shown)


MartinKolarik:
So are we sure these were mistagged and it wasn't a matching issue? Because I didn't see such probes at least in the current probe list.

alexey-yarmosh:
This PR targets these cases:

image
image

MartinKolarik:
Ok, seems valid then. They were not online previously, so I wasn't sure if you validated this was the right thing to fix (could have been a bug in probe matching, too).

MartinKolarik (approved):

PR #857, opened 2026-06-28
feat: credits farming anti abuse

### LINKED ISSUE #459: Credits farming anti abuse
Credits farming, e.g. starting 500 probes with some super cheap hosting provider. 

I would add rules like limiting the number of probes from the same ASN and same ASN+same city.

Maybe also a global probe limit per account

_Originally posted by @jimaek in https://github.com/jsdelivr/globalping/issues/452#issuecomment-1854061904_

alexey-yarmosh:
Fixes https://github.com/jsdelivr/globalping/issues/459

- allowing max 2 probes per user+asn+city
- allowing alt IPs in the same range
- primary IP should always has uniq IP range:
  - another primary IP in this range => disconnect to keep 1 primary IP.
  - alt IP in this range => disconnect primary IP, alt IP stays.

### DIFF
--- a/src/lib/ws/helper/probe-ip-limit.ts
+++ b/src/lib/ws/helper/probe-ip-limit.ts
@@ -1,20 +1,63 @@
-import _ from 'lodash';
 import config from 'config';
+import ipaddr from 'ipaddr.js';
+import { LRUCache } from 'lru-cache';
 import { scopedLogger } from '../../logger.js';
 import { ProbeError } from '../../probe-error.js';
+import type { ServerProbe, SocketProbe } from '../../../probe/types.js';
 import type { IoContext } from '../../server.js';
+import type { AdoptedProbes } from '../../override/adopted-probes.js';
+import type { AdoptionToken } from '../../../adoption/adoption-token.js';
 
 const numberOfProcesses = config.get<number>('server.processes');
+const asnCityPerUser = config.get<number>('probeLimit.asnCityPerUser');
 
 const logger = scopedLogger('ws:limit');
 
+const ipKeyCache = new LRUCache<string, string>({ max: 500_000 });
+
+export const getIpKey = (ip: string): string => {
+	const cached = ipKeyCache.get(ip);
+
+	if (cached !== undefined) {
+		return cached;
+	}
+
+	const parsed = ipaddr.parse(ip);
+	let ipKey = ip;
+
+	if (parsed.kind() === 'ipv6') {
+		const ipv6 = parsed as ipaddr.IPv6;
+		ipKey = ipv6.isIPv4MappedAddress()
+			? ipv6.toIPv4Address().toString()
+			: `${ipv6.parts.slice(0, 4).map(part => part.toString(16)).join(':')}::/64`;
+	}
+
+	ipKeyCache.set(ip, ipKey);
+
+	return ipKey;
+};
+
+const asnCityKey = (userId: string, location: { asn: number; city: string }) => JSON.stringify([ userId, location.asn, location.city ]);
+
+const addToSet = <K>(map: Map<K, Set<string>>, key: K, value: string) => {
+	const set = map.get(key) ?? new Set<string>();
+	set.add(value);
+	map.set(key, set);
+};
+
+const lowestSocketId = (socketIds: string[]) => socketIds.reduce((min, id) => id < min ? id : min);
+
+type UserProbe = Pick<ServerProbe, 'adoptionToken' | 'ipAddress' | 'uuid'> & { owner?: { id: string } };
+
 export class ProbeIpLimit {
 	private timer: NodeJS.Timeout | undefined;
+	private ipKeyIndexPromise: Promise<Map<string, Set<string>>> | undefined;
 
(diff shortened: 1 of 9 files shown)


src/probe/builder.ts
```diff
@@ -41,16 +43,14 @@ export const buildProbe = async (socket: Socket, probeIpLimit: ProbeIpLimit): Pr
 		throw new Error(`couldn't detect probe location for ip ${ip}`);
 	}
 
-	await probeIpLimit.verifyIpLimit(ip, socket.id);
-
```
MartinKolarik, line 45: This was moved to save lookup queries?

src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -1,73 +1,267 @@
...
 import type { IoContext } from '../../server.js';
+import type { AdoptedProbes } from '../../override/adopted-probes.js';
+import type { AdoptionToken } from '../../../adoption/adoption-token.js';
 
 const numberOfProcesses = config.get<number>('server.processes');
+const asnCityPerUser = config.get<number>('probeLimit.asnCityPerUser');
 
 const logger = scopedLogger('ws:limit');
 
+const ipKeyCache = new LRUCache<string, string>({ max: 500_000 });
```
MartinKolarik: 100k should be more than enough here?

src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -1,73 +1,267 @@
...
+			? ipv6.toIPv4Address().toString()
+			: `${ipv6.parts.slice(0, 4).map(part => part.toString(16)).join(':')}::/64`;
+	}
+
+	ipKeyCache.set(ip, ipKey);
+
+	return ipKey;
+};
+
+const asnCityKey = (userId: string, location: { asn: number; city: string }) => JSON.stringify([ userId, location.asn, location.city ]);
```
MartinKolarik: here a simple concat with `-` as separator should be fine

src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -1,73 +1,267 @@
...
-			this.syncIpLimit()
+			this.syncProbeLimits()
 				.finally(() => this.scheduleSync())
-				.catch(error => logger.error('Error in ProbeIpLimit.syncIpLimit()', error));
+				.catch(error => logger.error('Error in ProbeIpLimit.syncProbeLimits()', error));
 		}, 60_000 * 2 * Math.random() * numberOfProcesses).unref();
 	}
 
-	async syncIpLimit () {
+	async syncProbeLimits () {
```
MartinKolarik, line 74: This was actually fully sync and risked blocking the event loop. In my local runs, it was relatively fast (~20ms), but combined with other load on the server and with the probe list always growing, we should be more careful about stuff like this going forward (probably in many parts of the existing code too, but need to start somewhere). I added two helpers that can break the event loop in between more expensive function calls and within loops.

src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -1,73 +1,267 @@
...
+	}
+
+	private getUserId (probe: UserProbe): string | null {
+		const userId = probe.adoptionToken && this.adoptionToken.getUserIdByToken(probe.adoptionToken);
+
+		if (userId) {
+			return userId;
+		}
+
+		const dProbe = this.adoptedProbes.getByUuid(probe.uuid) || this.adoptedProbes.getByIp(probe.ipAddress);
```
MartinKolarik, line 265: This ignores alt IPs intentionally with the idea that they'll later get promoted to primary if the address changes and then rejected during the next scheduled run?

src/probe/builder.ts
alexey-yarmosh: Yes

src/lib/ws/helper/probe-ip-limit.ts
alexey-yarmosh: Yes, there's still plenty of room to spare.

src/lib/ws/helper/probe-ip-limit.ts
alexey-yarmosh: Yes, just used ':' as ids and cities use '-'.

src/lib/ws/helper/probe-ip-limit.ts
alexey-yarmosh: Nice, I'll keep that in mind in future 👍

src/lib/ws/helper/probe-ip-limit.ts
alexey-yarmosh: `adopted-probes.ts` will match the probe by alt IP and assign a user to it. So on the next run it will return the value by `getByIp(probe.ipAddress)`. 

We can update `getUserId` to handle alt IPs to disconnect immediately instead, but the case is too rare and I don't like the iteration over every alt IP on every `indexAsnCity` call for that.

MartinKolarik (approved):

PR #863, opened 2026-07-09
perf: cache credits reads and writes

alexey-yarmosh:
1. Delays writes to write max once per second per worker per user. 1 sec looks like a reasonable window to handle bursts, no point in making it bigger.
2. User credits >10000 threshold are cached for 60 sec with update on every write.
3. GET /v1/limits always gets fresh value from DB.


src/lib/credits.ts
```diff
@@ -1,20 +1,100 @@
...
+	async flush (): Promise<void> {
+		await Bluebird.map([ ...this.buffer.entries() ].filter(([ , entry ]) => entry.pending), async ([ userId, entry ]) => {
+			const flushed = entry.pending;
+			entry.pending = 0;
+
+			try {
+				await this.sql(CREDITS_TABLE).where({ user_id: userId }).update({ amount: this.sql.raw('amount - ?', [ flushed ]) });
+			} catch (error) {
+				if ((error as Error & { errno?: number }).errno === ER_CONSTRAINT_FAILED_CODE) {
+					logger.warn(`Dropped ${flushed} buffered credits of user ${userId} that exceed the current balance.`);
```
MartinKolarik: This rejects the whole update, so the user keeps even the credits he still has. Better (though we lose the error reporting):

```
amount: this.sql.raw('GREATEST(amount - ?, 0)', [ flushed ]),
```

MartinKolarik (commented):
Idea for an additional improvement: handle the buffer in the master process and have the workers call consume via a small promise-based wrapper over `process.send()` API. Fewer DB queries but also more precise (though still not 100%) info returned in the headers.

Also we need to update docs to say that the header info is approximate.

src/lib/ipc/ipc-master.ts
```diff
@@ -0,0 +1,52 @@
...
+		worker.send({ type: 'res', target, id, error: (error as Error).message });
+	}
+};
+
+export const initIPC = (): void => {
+	if (!cluster.isPrimary) {
+		throw new Error('`initIPC` called on non-primary process.');
+	}
+
+	cluster.on('fork', worker => worker.on('error', error => logger.error('Worker IPC channel error.', error)));
```
MartinKolarik: This seems like it would apply to any error, not just IPC.

src/lib/ipc/ipc-worker.ts
```diff
@@ -0,0 +1,56 @@
+import process from 'node:process';
+import type { IpcResponse } from './ipc-master.js';
+
+const IPC_TIMEOUT = 5000;
```
MartinKolarik: We should increase this to 10s and reduce `acquireConnectionTimeout` for dash to 5s, as it already is for other DBs, to minimize the chance this rejects while the credits are still deducted in the end.

MartinKolarik (approved):

PR #866, opened 2026-07-10
feat: do not wait for date_last_used update

alexey-yarmosh:


MartinKolarik (commented):
Not sure if this is worth the complications with undeterministic timing (e.g. in tests) when it only applies once per day.

alexey-yarmosh:
The effect is indeed too low, closing.

PR #867, opened 2026-07-13
fix: use fresh promise to build ipKey index

alexey-yarmosh:


src/lib/ws/helper/probe-ip-limit.ts
```diff
@@ -170,7 +173,6 @@ export class ProbeIpLimit {
 		for (const other of probes) {
 			if (other.location.asn !== probe.location.asn
 				|| other.location.city !== probe.location.city
-				|| other.client === probe.client
```
alexey-yarmosh, line 173: @MartinKolarik maybe I forgot about some specific case, but client/socketId is always a new value on every reconnect, so `other.client === probe.client` could never be true. Same on line 150.

MartinKolarik: Here I *think* it doesn't matter, but it would if it were called after `connect` to avoid matching the already-connected probe with itself.

alexey-yarmosh: Yes, middlewares are handled before the connect, so I'd remove the checks to avoid misleading that same probe is possible in `verify...` function.

MartinKolarik (approved):

PR #875, opened 2026-07-21
perf: reduce event loop delays

alexey-yarmosh:
Currently `syncPull` is the reason of ~600ms block every 2 seconds:
1. Optimising `syncPull` to call `updateProbes` per call instead of per node.
2. Optimising `updateProbes` itself.

### DIFF
--- a/src/lib/location/location.ts
+++ b/src/lib/location/location.ts
@@ -115,31 +115,30 @@ export const getRegionAliases = (key: string): string[] => {
 	return array ?? [];
 };
 
-const toIndexString = (s: string) => s.toLowerCase().replaceAll('-', ' ');
-const toIndexStrings = (array: string[]) => array.map(toIndexString);
+const toIndexStrings = (array: string[]) => array.map(s => s.toLowerCase().replaceAll('-', ' '));
 
 const getCountryCategories = _.memoize((country: string) => [
-	[ toIndexString(country) ],
-	[ toIndexString(getCountryIso3ByIso2(country)) ],
-	[ toIndexString(getCountryByIso(country)) ],
+	toIndexStrings([ country ]),
+	toIndexStrings([ getCountryIso3ByIso2(country) ]),
+	toIndexStrings([ getCountryByIso(country) ]),
 	toIndexStrings(getCountryAliases(country)),
 ]);
 
-const getCityCategory = _.memoize((normalizedCity: string) => [ toIndexString(normalizedCity) ]);
+const getCityCategory = _.memoize((normalizedCity: string) => toIndexStrings([ normalizedCity ]));
 
 const getStateCategories = _.memoize((state: string) => [
-	[ toIndexString(state) ],
-	[ toIndexString(getStateExtendedIsoByIso(state)) ],
-	[ toIndexString(getStateNameByIso(state)) ],
+	toIndexStrings([ state ]),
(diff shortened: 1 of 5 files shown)


src/lib/ws/synced-probe-list.ts
```diff
@@ -400,8 +404,9 @@ export class SyncedProbeList extends EventEmitter {
 		}
 
 		const eventsByNode = _.groupBy(eventsToProcess, event => event.message[MESSAGE_TYPES.NODE]);
+		let appliedUpdates = false;
```
MartinKolarik:
```suggestion
		let nodeDataChanged = false;
```

The current name is confusing, it's not "updated were applied" but "updates need to be applied"

src/lib/ws/synced-probe-list.ts
```diff
@@ -487,9 +493,20 @@ export class SyncedProbeList extends EventEmitter {
...
+		if (appliedUpdates) {
+			this.updateProbes();
+		}
+
+		const error = results.find(result => result.status === 'rejected');
+
+		if (error) {
+			throw error.reason;
+		}
+
```
MartinKolarik: This seems like a complicated way to write

```js
.finally(() => {
    if (appliedUpdates) {
	    this.updateProbes();
	}
})
```

src/lib/location/location.ts
```diff
@@ -115,27 +115,52 @@ export const getRegionAliases = (key: string): string[] => {
...
-		location.state ? [ getStateNameByIso(location.state) ] : [],
-		[ location.continent ],
-		location.continent ? [ getContinentName(location.continent) ] : [],
-		[ location.region ],
-		getRegionAliases(location.region),
+		...getCountryCategories(location.country),
+		getCityCategory(location.normalizedCity),
+		...(location.state ? getStateCategories(location.state) : [ [], [], [] ]),
+		...getContinentCategories(location.continent),
+		...getRegionCategories(location.region),
```
MartinKolarik: I don't like the spreads here much, given that we rely on specific index positions in other places, which makes it much harder to see what's where. Maybe destructure above into individual vars and keep here one var per line?

src/lib/location/location.ts
```diff
@@ -115,27 +115,52 @@ export const getRegionAliases = (key: string): string[] => {
...
 };
 
+const toIndexString = (s: string) => s.toLowerCase().replaceAll('-', ' ');
+const toIndexStrings = (array: string[]) => array.map(toIndexString);
+
+const getCountryCategories = _.memoize((country: string) => [
+	[ toIndexString(country) ],
+	[ toIndexString(getCountryIso3ByIso2(country)) ],
+	[ toIndexString(getCountryByIso(country)) ],
+	toIndexStrings(getCountryAliases(country)),
```
MartinKolarik: I'd use the array version everywhere to make it easier to read `toIndexStrings([ country ])`

src/lib/ws/synced-probe-list.ts
alexey-yarmosh: If handling of first node throws, we immediately get to `finally`, before other workers call `setNodeData`. So `allSettled` is used to call `setNodeData` for every worker before `updateProbes`, then throw an error if there were any.

MartinKolarik (approved):

PR #879, opened 2026-07-23
feat: add failureSource

MartinKolarik:
Adds a `failureSource` field to failed measurement results, allowing clients to distinguish target, resolver, and internal failures.

### DIFF
--- a/public/v1/components/schemas.yaml
+++ b/public/v1/components/schemas.yaml
@@ -1310,11 +1310,11 @@ components:
         - resolver
         - internal
       description: |
-        The source where the failure most likely occurred:
-         - `target` covers missing DNS records and protocol errors or timeouts,
+        The most likely source of the failure:
+         - `target` covers the target and its network path, including misconfigured DNS records and timeouts,
          - `resolver` covers failures communicating with or receiving an unusable response from the DNS resolver,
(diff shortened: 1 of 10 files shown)


public/v1/components/schemas.yaml
```diff
@@ -1301,6 +1303,18 @@ components:
...
+      type: string
+      enum:
+        - target
+        - resolver
+        - internal
+      description: |
+        The source where the failure most likely occurred:
+         - `target` covers missing DNS records and protocol errors or timeouts,
+         - `resolver` covers failures communicating with or receiving an unusable response from the DNS resolver,
+         - `internal` covers probe or API failures and policy rejections.
```
alexey-yarmosh: As I see, the field is meant to be used by services to distinguish between target failures and failures on GP side. So I think we need to add more guidance on how to use it, e.g.:
  ```
           - `target` covers missing DNS records and protocol errors or timeouts. Means something is likely wrong with the target itself.
           - `resolver` covers failures communicating with or receiving an unusable response from the DNS resolver. Means something is likely wrong with the used DNS resolver rather than the target.
           - `internal` covers probe or API failures and policy rejections. Means something is likely wrong on Globalping side and the measurement can be retried.
  ```

alexey-yarmosh (approved):

PR #882, opened 2026-07-27
feat: add measurement timeout option

### LINKED ISSUE #746: Timeout per measurement
Lets add a new standardized optional global parameter to all measurement types. Something like `timeout: Seconds`
- It would change the timeout per test type by changing the parameters we pass to the tools on the probe side
- Each test must exit cleanly without zombie processes being stuck. e.g. if I set a timeout of 10s and the server takes 50s to respond, the process on the probe must exit at 10s or be killed and the API must return "finished". 
- Return appropriate errors when timeouts are hit

MartinKolarik:
Closes https://github.com/jsdelivr/globalping/issues/746. The error classification is based on https://github.com/jsdelivr/globalping/pull/879 rather than adding a new field. The default timeout is now lower and based on what is realistically needed with the options we use.

### DIFF
--- a/src/measurement/schema/utils.ts
+++ b/src/measurement/schema/utils.ts
@@ -11,7 +11,7 @@ import {
 } from '../../lib/malware/client.js';
 import { joiValidate as joiMalwareValidateIp } from '../../lib/malware/ip.js';
 import { joiValidate as joiMalwareValidateDomain } from '../../lib/malware/domain.js';
-import type { MeasurementRecord, MeasurementRequest } from '../types.js';
+import type { MeasurementRecord, MeasurementRequest, RequestType } from '../types.js';
 import { isIpPrivate } from '../../lib/private-ip.js';
 
 export const globalIpOptions: { version: string[]; cidr: PresenceMode } = { version: [ 'ipv4', 'ipv6' ], cidr: 'forbidden' };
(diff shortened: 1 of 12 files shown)


src/schedule/executor.ts
```diff
@@ -128,6 +129,10 @@ export class StreamScheduleExecutor {
 				continue;
 			}
 
+			const timeout = getMeasurementTimeout(
+				configuration.measurement_type,
+				'packets' in configuration.measurement_options ? configuration.measurement_options.packets : undefined,
+			);
```
alexey-yarmosh: Since there are 3 calls of `getMeasurementTimeout`, maybe we can simplify its usage to be one line?
```
const timeout = getMeasurementTimeout(configuration.measurement_type, configuration.measurement_options);
```

test/tests/unit/measurement/schema/request-schema.test.ts
```diff
@@ -8,19 +8,62 @@ import {
...
+import type { RequestType } from '../../../../../src/measurement/types.js';
 import { populateDomainList, populateIpList } from '../../../../utils/populate-static-files.js';
 
+const withTimeout = <T extends { type: string; measurementOptions: object }>(request: T) => {
+	const packets = 'packets' in request.measurementOptions && typeof request.measurementOptions.packets === 'number'
+		? request.measurementOptions.packets
+		: undefined;
+
+	return { ...request, timeout: getMeasurementTimeout(request.type as RequestType, packets) };
+};
```
alexey-yarmosh: This util is harmful, since it uses the same function we are checking. Lets write expected value explicitly.

src/measurement/schema/utils.ts
```diff
@@ -143,6 +143,21 @@ export const COMMAND_DEFAULTS = {
...
+		return Math.max(5, Math.ceil((packets - 1) * 0.5 + 3));
+	}
+
+	if (type === 'mtr') {
+		packets ??= COMMAND_DEFAULTS.mtr.packets;
+		// Packet intervals + MTR grace + target DNS + ASN lookups.
+		return Math.max(5, Math.ceil(packets * 0.5 + 3 + 3 + 3));
+	}
+
+	return { http: 10, traceroute: 5, dns: 7 }[type];
```
alexey-yarmosh: For dns with `trace: true` the default of 7 seconds looks too low. Each queried server gets 2 tries of 3 seconds, so a single unresponsive server costs 6 seconds before dig switches to another one and still finishes successfully. E.g. here is a successful 10 second test: https://api.globalping.io/v1/measurements/2zhOg0Zi5IlmSVifv00020q4q. So for `trace: true` we probably want to add extra seconds to the default timeout. 

Also `+timeout=${Math.floor(options.timeout / 2)}` on gp-probe side doesn't fit `trace: true` - it makes every dead server proportionally more expensive as the timeout grows, fixed `+timeout=3` suits that case better.

alexey-yarmosh (changes requested):

src/measurement/schema/utils.ts
MartinKolarik: Good catch, I'll make the timeout fixed for trace and increase the default.

alexey-yarmosh (approved):

PR #893, opened 2026-08-10
feat: sync probe settings

PavelKopecky:
Needed for https://github.com/jsdelivr/globalping-dash/issues/165

### DIFF
--- a/migrations/dashboard/create-tables.js.sql
+++ b/migrations/dashboard/create-tables.js.sql
@@ -46,6 +46,10 @@ CREATE TABLE IF NOT EXISTS gp_probes (
 	originalLocation LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`originalLocation`)),
 	customLocation LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`customLocation`)),
 	localAdoptionServer LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`localAdoptionServer`)),
+	settings LONGTEXT COLLATE utf8mb4_bin
+		DEFAULT '{"meteredConnection":false}'
+		NOT NULL
+		CHECK (JSON_VALID(`settings`)),
 	searchIndex VARCHAR(4090) NULL
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 
--- a/src/lib/override/adopted-probes.ts
+++ b/src/lib/override/adopted-probes.ts
@@ -349,12 +349,7 @@ export class AdoptedProbes {
 		const probes = this.getProbesWithAdminData();
 		// 'probe' - usual API probe. 'dProbe' - dashboard probe data stored in sql.
 		const { dProbesWithProbe, dProbesWithoutProbe, probesWithoutDProbe } = this.matchDProbesAndProbes(probes);
-
-		for (const { dProbe, probe } of dProbesWithProbe) {
-			if (!_.isEqual(probe.settings, dProbe.settings)) {
-				this.emitToProbe(probe.client, 'api:settings:update', dProbe.settings);
-			}
-		}
+		this.syncProbeSettings(dProbesWithProbe);
 
 		const { updatedDProbes, dProbeDataUpdates } = this.generateUpdatedDProbes(dProbesWithProbe, dProbesWithoutProbe);
 		const { dProbesToDelete, nullifyIpUpdates, dProbeAltIpUpdates } = this.findDProbeDuplicates(updatedDProbes);
@@ -603,6 +598,14 @@ export class AdoptedProbes {
 
 		const probesWithoutDProbe = [ ...uuidToProbe.values() ];
 		return { dProbesWithProbe, dProbesWithoutProbe, probesWithoutDProbe };
+	}
+
+	private syncProbeSettings (dProbesWithProbe: { dProbe: DProbe; probe: SocketProbe }[]) {
+		for (const { dProbe, probe } of dProbesWithProbe) {
+			if (!_.isEqual(probe.settings, dProbe.settings)) {
+				this.emitToProbe(probe.client, 'api:settings:update', dProbe.settings);
+			}
+		}
 	}
 
 	private generateUpdatedDProbes (dProbesWithProbe: { dProbe: DProbe; probe: SocketProbe }[], dProbesWithoutProbe: DProbe[]) {
--- a/src/lib/override/adopted-probes.ts
+++ b/src/lib/override/adopted-probes.ts
@@ -126,7 +126,6 @@ export class AdoptedProbes {
 	private idToDProbe: Map<string, DProbe> = new Map();
 	private ipToDProbe: Map<string, DProbe> = new Map();
 	private uuidToDProbe: Map<string, DProbe> = new Map();
-	private settingsUpdateHandler?: (probe: SocketProbe) => void;
 	private syncBackToDashboard = process.env['SHOULD_SYNC_ADOPTIONS'] === 'true';
 	static readonly dProbeFieldToProbeField = {
 		uuid: {
@@ -247,7 +246,7 @@ export class AdoptedProbes {
 	constructor (
 		private readonly sql: Knex,
 		private readonly getProbesWithAdminData: () => SocketProbe[],
-		private readonly fetchLocalSocketProbes: () => Promise<SocketProbe[]>,
+		private readonly emitToProbe: (client: string, event: string, payload: unknown) => void,
 	) {}
 
 	getById (id: string) {
@@ -260,10 +259,6 @@ export class AdoptedProbes {
 
 	getByUuid (uuid: string) {
 		return this.uuidToDProbe.get(uuid) || null;
-	}
-
-	setSettingsUpdateHandler (handler: (probe: SocketProbe) => void) {
-		this.settingsUpdateHandler = handler;
 	}
 
 	getUpdatedLocation (probe: SocketProbe, adminLocation?: ProbeLocation | null): ExtendedProbeLocationWithOverrides | null {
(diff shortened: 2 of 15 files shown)


src/lib/override/adopted-probes.ts
```diff
@@ -395,6 +401,7 @@ export class AdoptedProbes {
 			customLocation: row.customLocation ? JSON.parse(row.customLocation) as DProbe['customLocation'] : null,
 			originalLocation: row.originalLocation ? JSON.parse(row.originalLocation) as DProbe['originalLocation'] : null,
 			localAdoptionServer: row.localAdoptionServer ? JSON.parse(row.localAdoptionServer) as DProbe['localAdoptionServer'] : null,
+			settings: JSON.parse(row.settings) as ProbeSettings,
```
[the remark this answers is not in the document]
PavelKopecky: ProbeSettings is validated in Directus

migrations/dashboard/create-tables.js.sql
```diff
@@ -46,6 +46,10 @@ CREATE TABLE IF NOT EXISTS gp_probes (
 	originalLocation LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`originalLocation`)),
 	customLocation LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`customLocation`)),
 	localAdoptionServer LONGTEXT COLLATE utf8mb4_bin NULL CHECK (json_valid(`localAdoptionServer`)),
+	settings LONGTEXT COLLATE utf8mb4_bin
+		DEFAULT '{"meteredConnection":false}'
+		NOT NULL
+		CHECK (JSON_VALID(`settings`)),
```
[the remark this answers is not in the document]
PavelKopecky: Similar edits have been made across 35 other commits; test script drops all tables and reruns the migration, dev may need to be recreated, as it always has -- that's ok.

src/lib/override/adopted-probes.ts
```diff
@@ -30,6 +30,10 @@ export const escapeMdSymbols = (value: string): string => value
 	.replace(/_/g, '\\_')
 	.replace(/`/g, '\\`');
 
+export type ProbeSettings = {
+	meteredConnection?: boolean;
```
alexey-yarmosh: ? should not be required.

src/lib/override/adopted-probes.ts
```diff
@@ -86,6 +90,7 @@ type DProbe = {
 		expiresAt: string;
 		ips: string[];
 	} | null;
+	settings: ProbeSettings;
```
alexey-yarmosh: Creation of an extra type is not necessary, use `settings: { meteredConnection?: boolean; }` and `as DProbe['settings']`.

alexey-yarmosh (changes requested):

src/lib/override/adopted-probes.ts
```diff
@@ -395,6 +399,7 @@ export class AdoptedProbes {
 			customLocation: row.customLocation ? JSON.parse(row.customLocation) as DProbe['customLocation'] : null,
 			originalLocation: row.originalLocation ? JSON.parse(row.originalLocation) as DProbe['originalLocation'] : null,
 			localAdoptionServer: row.localAdoptionServer ? JSON.parse(row.localAdoptionServer) as DProbe['localAdoptionServer'] : null,
+			settings: JSON.parse(row.settings) as DProbe['settings'],
```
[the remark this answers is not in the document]
PavelKopecky: it is a required field

src/lib/override/adopted-probes.ts
PavelKopecky: https://github.com/jsdelivr/globalping/pull/893/commits/57b9430cc768bdbe80511ca9982d69e43b8e3ac6

src/lib/override/adopted-probes.ts
PavelKopecky: https://github.com/jsdelivr/globalping/pull/893/changes/b1dfb6cc3ff0307ee524db86ccd575ef29475b0e

src/lib/ws/gateway.ts
```diff
@@ -30,6 +30,9 @@ export const initGateway = (ioContext: IoContext) => {
 			const probe = socket.data.probe;
 			const location = probeOverride.getUpdatedLocation(probe);
 
+			const settings = adoptedProbes.getByIp(probe.ipAddress)?.settings;
+			settings && socket.emit('api:settings:update', settings);
```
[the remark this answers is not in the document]
PavelKopecky: This is intended and matches https://github.com/jsdelivr/globalping/blob/master/src/adoption/adoption-token.ts#L79

alexey-yarmosh:
When you are done re-request the review please.

PavelKopecky:
@alexey-yarmosh after a discussion with @.MartinKolarik we've decided to persist the settings on the probe side as well; see https://github.com/jsdelivr/globalping-probe/pull/405... so there's more changes

src/lib/override/adopted-probes.ts
```diff
@@ -336,6 +346,16 @@ export class AdoptedProbes {
...
+		const localSocketProbes = await this.fetchLocalSocketProbes();
+
+		for (const probe of localSocketProbes) {
+			const dProbe = this.getByUuid(probe.uuid);
+
+			if (dProbe && !_.isEqual(probe.settings, dProbe.settings)) {
+				probe.settings = dProbe.settings;
+				this.settingsUpdateHandler?.(probe);
+			}
+		}
```
alexey-yarmosh: Instead of dealing with local probes and mutating them here and in the `gateway.ts`. Lets:
1. Single worker (the one with `syncBackToDashboard: true`) takes `dProbesWithProbe` and emits `api:settings:update` to all probes probes with outdated settings, not only local.
2. Probe does the settings update and emits `probe:settings:update` which is handled by the worker it is connected to, which does the mutation.
3. Emit in `gateway.ts` stays.

src/lib/server.ts
```diff
@@ -76,18 +77,28 @@ export const createServer = async ({ startBackgroundJobs = true }: CreateServerO
...
 
 	const getProbesWithAdminData = (): SocketProbe[] => syncedProbeList.getProbesWithAdminData();
-	const adoptedProbes = new AdoptedProbes(dashboardClient, getProbesWithAdminData);
+
+	const fetchRawLocalProbes = async () => {
+		const rawSockets = await ioContext.fetchRawLocalSockets();
+		return rawSockets.map(socket => socket.data.probe);
+	};
+
+	const adoptedProbes = new AdoptedProbes(dashboardClient, getProbesWithAdminData, fetchRawLocalProbes);
```
alexey-yarmosh: Instead of `setSettingsUpdateHandler` lets do:
```ts
const emitToProbe = (client: string, event: string, payload: unknown) => {
		io.of(PROBES_NAMESPACE).to(client).emit(event, payload);
	};

	const adoptedProbes = new AdoptedProbes(dashboardClient, getProbesWithAdminData, emitToProbe);
```

alexey-yarmosh (changes requested):

src/lib/override/adopted-probes.ts
MartinKolarik: I'm not quite sure about this. The initial idea was that the server (based on Directus) would always control the settings, so the flow would be one-way. Adding the emit back from the probe to avoid direct assignment here seems unnecessary. Yes, it matches how probe reports other things, but those are probe-controlled, while the settings are not.

alexey-yarmosh: The reason is simpler design, as right now we need to mutate the probe in every place we are sending the settings, which looks messy and is easy to forget (like in gateway.ts). Additionally, it gives ack from probe and retries if no ack.

PavelKopecky: https://github.com/jsdelivr/globalping/pull/893/commits/915f5f0900e9c6f1b306ee9610cbfa7727145441

src/lib/server.ts
PavelKopecky: https://github.com/jsdelivr/globalping/pull/893/commits/915f5f0900e9c6f1b306ee9610cbfa7727145441

src/lib/override/adopted-probes.ts
```diff
@@ -344,6 +349,13 @@ export class AdoptedProbes {
 		const probes = this.getProbesWithAdminData();
 		// 'probe' - usual API probe. 'dProbe' - dashboard probe data stored in sql.
 		const { dProbesWithProbe, dProbesWithoutProbe, probesWithoutDProbe } = this.matchDProbesAndProbes(probes);
+
+		for (const { dProbe, probe } of dProbesWithProbe) {
+			if (!_.isEqual(probe.settings, dProbe.settings)) {
+				this.emitToProbe(probe.client, 'api:settings:update', dProbe.settings);
+			}
+		}
```
alexey-yarmosh: Move this to a new method please.

alexey-yarmosh (changes requested):

src/lib/override/adopted-probes.ts
PavelKopecky: https://github.com/jsdelivr/globalping/pull/893/commits/ee0c26f7861ca31788824110491ff7f906de0060

src/lib/ws/gateway.ts
```diff
@@ -30,6 +31,9 @@ export const initGateway = (ioContext: IoContext) => {
 			const probe = socket.data.probe;
 			const location = probeOverride.getUpdatedLocation(probe);
 
+			const settings = adoptedProbes.getByIp(probe.ipAddress)?.settings;
```
alexey-yarmosh:
```suggestion
			const settings = (adoptedProbes.getByUuid(probe.uuid) || adoptedProbes.getByIp(probe.ipAddress))?.settings;

```

alexey-yarmosh (changes requested):

alexey-yarmosh (approved):

MartinKolarik (approved):

PR #896, opened 2026-08-18
feat!: move probe logs to timescale db

PavelKopecky:
API part of https://github.com/jsdelivr/globalping-dash/issues/210

BREAKING:
**The `after` parameter now represents a probeLogId, not a Redis ID. This will require an update of the dashboard as well, and must be deployed together with that update.** 

see https://github.com/jsdelivr/globalping-dash/pull/218


test/e2e/cases/probe-logs.test.ts
```diff
@@ -136,33 +139,46 @@ describe('probe logs', () => {
...
 
 		const newResponse = await waitForLogSync(PROBE_ID, authCookie, lastId as string);
 
 		expect(newResponse.body.lastId).to.exist;
+		expect(newResponse.body.lastId).to.match(/^\d+$/);
 		expect(newResponse.body.lastId).to.not.equal(lastId);
 
 		const newTimestamps = newResponse.body.logs.map(log => log.timestamp);
-		expect(_.intersection(initialTimestamps, newTimestamps).length).to.equal(0);
+		expect(_.intersection(settledTimestamps, newTimestamps).length).to.equal(0);
```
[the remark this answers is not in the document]
PavelKopecky: timestamps will definitely be different before and after restart

PavelKopecky:
@coderabbitai full review

src/probe/handler/logs.ts
```diff
@@ -22,6 +22,12 @@ export const handleNewLogs = (probe: SocketProbe) => async (logMessage: LogMessa
 		throw validation.error;
 	}
 
-	await probeLogStorage.writeLogs(probe.uuid, logMessage);
+	try {
+		await probeLogStorage.writeLogs(probe.uuid, logMessage);
+	} catch (error: unknown) {
+		callback?.('error');
+		throw error;
```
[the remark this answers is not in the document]
PavelKopecky: Theres no good reason for the callback to throw.

src/probe/schema/get-probe-logs-schema.ts
```diff
@@ -1,5 +1,37 @@
...
+		|| (normalized.length === MAX_PROBE_LOG_ID.length && normalized > MAX_PROBE_LOG_ID)) {
+		return helpers.error('any.invalid');
+	}
+
+	return value;
+});
+
+export const schema = Joi.object<GetProbeLogsQuery>({
+	after: afterSchema.optional(),
+	scopes: scopesSchema.optional(),
```
alexey-yarmosh, line 37: Lets add length validation to after (20) and to scopes (1024).

migrations/time-series-1/20260815120000_create-probe-logs.js
```diff
@@ -0,0 +1,23 @@
...
+		table.bigInteger('probeLogId').notNullable();
+		table.timestamp('timestamp', { useTz: true });
+		table.timestamp('receivedAt', { useTz: true }).notNullable();
+		table.string('level', 8);
+		table.string('scope', 64);
+		table.string('message', 8192).notNullable();
+	});
+
+	await db.raw(`SELECT create_hypertable('probe_log', by_range('receivedAt', INTERVAL '1 day'))`);
+	await db.raw(`SELECT add_retention_policy('probe_log', INTERVAL '3 days')`);
```
alexey-yarmosh: Increasing retention to ~30 days might be helpful for users. For example, they can take a look at last logs of the probe which was disconnected 1 week ago. @MartinKolarik wdyt?

alexey-yarmosh:
Looks like `hasMore` and `nextCursor` part of the design is missing, so user can't go back in history further than 1000 lines.

alexey-yarmosh (changes requested):

migrations/time-series-1/20260815120000_create-probe-logs.js
MartinKolarik: I suppose we can increase the time limit, but not the count limit right now, that would be too much volume. I'll want to do that after https://github.com/jsdelivr/globalping-dash/issues/217

alexey-yarmosh: Yes, for the described case only time limit increase is required.

PavelKopecky:
> Looks like `hasMore` and `nextCursor` part of the design is missing, so user can't go back in history further than 1000 lines.

The original proposed design wouldn't have worked with the dashboard (no live tail support because of the missing `after` parameter) and the dashboard **currently** has no use for the `hasMore` and `nextCursor` params. Do we still want to add those? @MartinKolarik

migrations/time-series-1/20260815120000_create-probe-logs.js
PavelKopecky: https://github.com/jsdelivr/globalping/pull/896/commits/579d5b7bb37482ca304fa148156695a6dfe84bc3

src/probe/schema/get-probe-logs-schema.ts
PavelKopecky: https://github.com/jsdelivr/globalping/pull/896/changes/c598a02e6f1b16cc83fb49fc7fe7537b805d852d

src/probe/logs-storage.ts
```diff
@@ -1,51 +1,225 @@
...
+			WHERE "lastSeenAt" >= now() - ?::interval
+				AND "scope" !~ '(^[[:space:]]|[[:space:]]$|,)'
+			GROUP BY "scope"
+			HAVING count(*) >= ?
+			ORDER BY count(*) DESC, "scope" ASC
+			LIMIT ?
+		`, [ SCOPE_RETENTION, MIN_SCOPE_PROBES, SCOPE_OPTIONS_LIMIT ]);
+
+		return result.rows.map(({ scope }) => scope);
 	}
```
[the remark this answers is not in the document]
PavelKopecky: This is intentional from my side, but I am not fully sure whether that's actually the correct behavior. What do you think @alexey-yarmosh?

PavelKopecky:
Recent changes needed for https://github.com/jsdelivr/globalping-dash/pull/218#discussion_r3836603491

src/probe/logs-storage.ts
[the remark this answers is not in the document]
alexey-yarmosh: Yes, public endpoint is fine.

src/probe/logs-storage.ts
```diff
@@ -1,51 +1,225 @@
...
+import type { Knex } from 'knex';
+import type { LogMessage } from './handler/logs.js';
+import { timeSeriesClient } from '../lib/sql/client.js';
 
-const REDIS_ID_REGEX = /^\d+-\d+$/;
+const READ_LIMIT = 1000;
+const CLEANUP_INTERVAL = 10_000n;
+const MAX_RETAINED_LOGS = 100_000n;
+const MAX_SCOPES_PER_PROBE = 100;
+const MIN_SCOPE_PROBES = 2;
```
alexey-yarmosh: `MIN_SCOPE_PROBES` is a good idea but its not secure enough (scope from 2 patched probes) and accurate enough (rare scopes from 1 probe). Instead let probes report all scopes passed to `scopedLogger()`. Something like that:

**Suggestion: have the probe report its own scopes instead of mining them from logs.**

  Scopes aren't runtime data - they're 18 hardcoded `scopedLogger('...')` literals in the probe. Deriving them from logs can't work well: 12 of those 18 only log on error (`ping-command`, `mtr-command`, `dns-command`, `traceroute-command` each have nothing but `logger.error`), so any popularity threshold drops
  exactly the scopes you'd want to filter by. In 26h of a real probe's logs only 6 of 18 scopes ever appeared.

  Reporting them at connect fixes that and lets the threshold be a share of the fleet, which also closes the unauthenticated-endpoint concern - polluting the list would require controlling half the probes.

  **Probe** - all `scopedLogger()` calls are top-level and every command module is statically imported in `probe.ts`, so the set is complete by connect time:

  ```ts
  export const registeredScopes = new Set<string>();
  export const scopedLogger = (scope: string) => {
    registeredScopes.add(scope);
    return logger.child({ scope });
  };
  // on connect
  socket.emit('probe:log-scopes', [ ...registeredScopes ]);
  ```

  **Handler:**

  ```ts
  const logScopesSchema = Joi.array().max(64).items(Joi.string().max(64)).required();

  export const handleLogScopes = (probeUuid: string) => async (scopes: string[]) => {
    const { error, value } = logScopesSchema.validate(scopes);
    if (error) { throw error; }

    const redis = getPersistentRedisClient();

    await Promise.all([
      redis.sAdd('gp:log-scopes', value),
      ...value.flatMap(scope => [
        redis.sAdd(`gp:log-scope:${scope}`, probeUuid),
        // NX: never extend. Otherwise the key never expires, dead uuids accumulate forever,
        // and a newly added scope can never reach the threshold against an inflated total.
        redis.expire(`gp:log-scope:${scope}`, 10 * 24 * 3600, 'NX'),
      ]),
    ]);
  };
  ```

  **Read:**

  ```ts
  async readScopes (): Promise<string[]> {
    const redis = getPersistentRedisClient();
    const known = await redis.sMembers('gp:log-scopes');
    const counts = await Promise.all(known.map(s => redis.sCard(`gp:log-scope:${s}`)));
    const total = Math.max(0, ...counts);   // `general` is on every probe, so max ≈ fleet size

    return known.filter((_, i) => counts[i]! >= total * 0.5).sort();
  }
  ```

  Sets rather than counters, so reconnect floods can't inflate a scope - `SADD` of the same uuid is a no-op. 10-day TTL is above `uptime.maxUptime` (7d) + `maxDeviation` (1d), so every live probe repopulates its key at least once per window. No `multi()` needed - node-redis pipelines queued commands on the next
  tick anyway, and we don't need atomicity here.

  This drops `probe_log_scope`, its migration, the cleanup procedure and its daily job, the historic backfill, the 100-scope cap, and the per-flush upsert inside the log-write transaction. Nothing is written to Postgres on the hot path.
  
  cc @MartinKolarik

alexey-yarmosh (changes requested):

src/probe/logs-storage.ts
MartinKolarik: I like the idea, but the read part seems rather heavy. I'd consider alternative storage structures. E.g., based on the return of `sAdd` conditionally increment a `scope -> count` value (could be in a script).

alexey-yarmosh: Scopes should expire at some time, and its hard to sync it with counter. So a few updates:
- caching read result;
- We can't set TTL per scope because nothing triggers its refill after. Using `zAdd` + `zRemRangeByScore` for a sliding window instead.
- sending from probe shouldn't be immediate to spread the load;
- adding by `ip`, not by `uuid`.

**Probe** :

  ```ts
  export const registeredScopes = new Set<string>();
  export const scopedLogger = (scope: string) => {
    registeredScopes.add(scope);
    return logger.child({ scope });
  };

  // on connect
  setTimeout(() => socket.emit('probe:log-scopes', [ ...registeredScopes ]), _.random(0, 60_000));
  ```

  **Handler:**

  ```ts
  const WINDOW = 10 * 24 * 3600;

  const logScopesSchema = Joi.array()
    .max(64)
    .items(Joi.string().max(64).pattern(/^[a-zA-Z0-9:_-]+$/))
    .required();

  export const handleLogScopes = (probe: SocketProbe) => async (scopes: string[]) => {
    const { error, value } = logScopesSchema.validate(scopes);
    if (error) { throw error; }

    const redis = getPersistentRedisClient();
    const now = Math.floor(Date.now() / 1000);

    await Promise.all([
      redis.sAdd('gp:log-scopes', value),
      ...value.flatMap(scope => [
        redis.zAdd(`gp:log-scope:${scope}`, { score: now, value: probe.ipAddress }),
        redis.zRemRangeByScore(`gp:log-scope:${scope}`, '-inf', now - WINDOW),
        redis.expire(`gp:log-scope:${scope}`, WINDOW),
      ]),
    ]);
  };
  ```

  **Read:**

  ```ts
  const cache = new TTLCache<string, string[]>({ ttl: 3600_000 });

  async readScopes (): Promise<string[]> {
    const cached = cache.get('scopes');
    if (cached) { return cached; }

    const redis = getPersistentRedisClient();
    const known = await redis.sMembers('gp:log-scopes');
    const counts = await Promise.all(known.map(s => redis.zCard(`gp:log-scope:${s}`)));
    const threshold = Math.max(Math.max(0, ...counts) * 0.5, 10);

    const dead = known.filter((_, i) => counts[i] === 0);
    if (dead.length) { await redis.sRem('gp:log-scopes', dead); }

    const scopes = known.filter((_, i) => counts[i]! >= threshold).sort();
    cache.set('scopes', scopes);
    return scopes;
  }
  ```

src/probe/route/get-probe-log-scopes.ts
```diff
@@ -0,0 +1,15 @@
+import type { ExtendedContext, ExtendedRouter } from '../../types.js';
+import { getProbeLogScopesStorage } from '../log-scopes-storage.js';
+
+const probeLogScopesStorage = getProbeLogScopesStorage();
+
+export const registerGetProbeLogScopesRoute = (router: ExtendedRouter): void => {
+	const handle = async (ctx: ExtendedContext) => {
+		const scopes = await probeLogScopesStorage.readScopes();
```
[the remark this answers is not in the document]
PavelKopecky: I don't think that's necessary? @alexey-yarmosh

test/tests/integration/measurement/probe-communication.test.ts
```diff
@@ -7,7 +7,10 @@ import * as sinon from 'sinon';
 import { expect } from 'chai';
 import { ConsoleWriter } from 'h-logger2';
 import nockGeoIpProviders from '../../../utils/nock-geo-ip.js';
+import { waitFor } from '../../../utils/wait.js';
 import * as id from '../../../../src/measurement/id.js';
+import { getPersistentRedisClient } from '../../../../src/lib/redis/persistent-client.js';
+import { REPORTER_SCOPES_KEY_PREFIX, SCOPE_KEY_PREFIX } from '../../../../src/probe/log-scopes-storage.js';
```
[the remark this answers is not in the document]
PavelKopecky: This test reports from only one IP, so the scope never reaches the 10-reporter threshold required for insertion into KNOWN_SCOPES_KEY and therefore creates no entry to clean up.

src/probe/route/get-probe-log-scopes.ts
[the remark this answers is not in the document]
MartinKolarik: No.

alexey-yarmosh (approved):

src/probe/log-scopes-storage.ts
```diff
@@ -0,0 +1,78 @@
+import { getPersistentRedisClient, type RedisClient } from '../lib/redis/persistent-client.js';
+import { getIpKey } from '../lib/ws/helper/probe-ip-limit.js';
+
+export const KNOWN_SCOPES_KEY = 'gp:log-scopes';
+export const SCOPE_KEY_PREFIX = 'gp:log-scope:';
+export const REPORTER_SCOPES_KEY_PREFIX = 'gp:log-scopes:reporter:';
+export const SCOPE_ACTIVE_WINDOW = 864_000; // 10 days in seconds
+export const SCOPE_READ_CACHE_TTL = 3_600_000; // 1 hour in ms
+export const MIN_SCOPE_REPORTERS = 10;
```
MartinKolarik: Please make the minimum number of scope reporters configurable, keeping the default at 10 and overriding it to 2 in config/development.cjs. This lets scope reporting work with a small local probe fleet.

PR #903, opened 2026-08-31
feat: add reliability metrics

MartinKolarik:
Adds metrics for probe-sync delays and event volume, no-probe measurement failures, measurement completion and timeout outcomes, and persistence backlog depth.

Note that there's a small race between timeout tracking and completion when the results arrive near the timeout, I think it's acceptable and introducing a new redis data structure needed to avoid it is not worth it.

### DIFF
--- a/src/lib/metrics-collector.ts
+++ b/src/lib/metrics-collector.ts
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import { Queue as BullQueue } from 'bullmq';
 import { monitorEventLoopDelay } from 'node:perf_hooks';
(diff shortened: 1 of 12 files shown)


alexey-yarmosh (changes requested):
`executor.ts` doesn't call `metricsAgent.recordMeasurement(request.type, onlineProbesMap.size);` so scheduled measurement doesn't commit to `gp.measurement.count.*`, but still commits to `gp.measurement.completed.*` and `gp.measurement.timeout.*` on finish.

alexey-yarmosh (approved):

PR #914, opened 2026-09-10
feat: add HTTP request blacklist

MartinKolarik:
Rejects HTTP measurements that match an API-side request blacklist, initially covering sensitive file and directory path suffixes derived from [CrowdSec's sensitive-data list](https://hub-data.crowdsec.net/web/sensitive_data.txt).


src/measurement/schema/http-request-blacklist.ts
```diff
@@ -0,0 +1,42 @@
...
+	.split(/\r?\n/)
+	.map(value => value.trim())
+	.filter(Boolean)
+	.map(value => _.escapeRegExp(value));
+
+if (blacklistedPaths.length === 0) {
+	throw new Error('HTTP request blacklist is empty');
+}
+
+const blacklistedPathRegExp = new RegExp(`(?:^|/)(?:${blacklistedPaths.join('|')})/?$`);
```
alexey-yarmosh: It might make sense to add `i` to regex, so `/.ENV` is rejected as well as `/.env` since on some servers it's the same file.

alexey-yarmosh (changes requested):

alexey-yarmosh (approved):

